PR #39, opened 2024-07-16
Development of the concurrency control module for the new agent

jr0me:
### Description

This pull request introduces a foundational TaskManager class designed to meet the evolving needs of the Agent. The TaskManager can handle both thread-based and coroutine-based tasks, using Boost.Asio for asynchronous operations.

The class implements a generic interface, ITaskManager, allowing it to manage and execute tasks via an io_context. It supports enqueuing both standard function tasks and coroutine tasks, using boost::asio::co_spawn for the latter. This initial implementation serves as a starting point for developing a comprehensive task management system that will accommodate the Agent's future needs, integrating both threading and coroutine paradigms.


TomasTurina (approved):
LGTM!

Issue #26, opened 2024-07-02
Design and development of the configuration parser component for the new agent


TomasTurina:
**Parent issue:**
- https://github.com/wazuh/wazuh-agent/issues/14

## Description

As part of the development of the new agent [MVP](https://github.com/wazuh/wazuh-agent/issues/14), it is necessary to develop a new configuration parser module.

Take into account the following:

- Configuration will be in TOML format.
  - The name of the configuration file will be `wazuh.conf`.
- Avoid to create a new library for this.
  - Select and use a common and maintained library, mention the advantages and disadvantages of its use.
- This module should be able to parse and save the configuration in memory.
  - Consider STL containers to do this.
- This module should be able to distribute all configuration to all modules.
  - Each module will be in charge of reading and applying its particular configuration.

Nicogp:
## Libraries investigation

The official [TOML Wiki](https://github.com/toml-lang/toml/wiki) lists three C++ libraries:

- [toml11](https://github.com/ToruNiina/toml11)
- [tomlcpp](https://github.com/cktan/tomlcpp)
- [toml++](https://github.com/cktan/tomlcpp)

### toml11
[Documentation](https://toruniina.github.io/toml11/)
Recent activity in the repository.
Licence: MIT
Standar: C++11/14/17/20
Features:

- It complies with [the latest TOML language specification](https://toml.io/en/v1.0.0).
- It supports new features merged into the upcoming TOML version (v1.1.0).
- It provides clear error messages, including the location of the error.
- It parses and retains comments, associating them with corresponding values.
- It provides exception-less parse function.
- It supports complex type conversions from TOML values.
- It allows customization of the types stored in toml::value.


### tomlcpp
[Documentation](https://github.com/cktan/tomlcpp) (Only a Readme)
Last update: two years ago.
Licence: MIT
Standar: This is a C++ wrapper around the C library available here: https://github.com/cktan/tomlc99.
Features:

- Compatible with [TOML v1.0.0](https://toml.io/en/v1.0.0).

- Does not throw C++ exceptions.
- Provides very simple and intuitive interface.

### toml++
[Documentation](https://marzer.github.io/tomlplusplus/)
Last update: 4 months ago.
Licence: MIT
Standar: C++17 (plus some C++20 features where available, e.g. experimental support for [char8_t]
Features:

- Header-only (optional!)
[...]

PR #46, opened 2024-07-18
Configuration parser module mvp

Nicogp:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/26|

## Description
This PR adds the configurationParser module, it reads the configuration from a wazuh.conf file in toml format and then exposes a GetConfig<T>(keys) method to query the different values using one or more keys. 
It uses the TOML11 library, this library is the one that is receiving more updates, it is based on C++20 and supports the latest version of TOML, also its documentation is clear and with examples.

## Tests

  - [x] Linux

- [x] Added unit tests (for new features)

### DIFF
--- a/src/agent/configuration_parser/CMakeLists.txt
+++ b/src/agent/configuration_parser/CMakeLists.txt
@@ -8,35 +8,20 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
 set(CMAKE_TOOLCHAIN_FILE "${CMAKE_SOURCE_DIR}/../../vcpkg/scripts/buildsystems/vcpkg.cmake")
 set(VCPKG_MANIFEST_DIR ${CMAKE_SOURCE_DIR}/../../)
 
+project(ConfigurationParser)
+
 find_package(toml11 CONFIG REQUIRED)
 
-project(ConfigurationParser)
-message(STATUS "******* Project ConfigurationParser *********")
-message(STATUS "CMAKE_SOURCE_DIR: ${CMAKE_SOURCE_DIR}")
-message(STATUS "CMAKE_CURRENT_SOURCE_DIR: ${CMAKE_CURRENT_SOURCE_DIR}")
+add_library(ConfigurationParser src/configuration_parser.cpp)
 
-include_directories(include)
-include_directories(${toml11_INCLUDE_DIRS})
-message(STATUS "Include directories: ${toml11_INCLUDE_DIRS}")
-
+target_include_directories(ConfigurationParser PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_link_libraries(ConfigurationParser PUBLIC toml11::toml11)
 
 if(BUILD_TESTS)
     find_package(GTest CONFIG REQUIRED)
 
     enable_testing()
 
-    add_executable(ConfigurationParser_test
-        tests/configuration_parser_test.cpp
-        src/configuration_parser.cpp
-    )
-
-    target_include_directories(ConfigurationParser_test PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
-    target_link_libraries(ConfigurationParser_test PUBLIC toml11::toml11 GTest::gtest GTest::gtest_main GTest::gmock GTest::gmock_main)
+    add_executable(ConfigurationParser_test tests/configuration_parser_test.cpp)
+    target_link_libraries(ConfigurationParser_test PUBLIC ConfigurationParser GTest::gtest GTest::gtest_main GTest::gmock GTest::gmock_main)
 endif()
-
-add_library(configurationParser src/configuration_parser.cpp)
-
-set_target_properties(configurationParser PROPERTIES LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
-
-target_include_directories(configurationParser PUBLIC)
-target_link_libraries(configurationParser PUBLIC toml11::toml11)
--- a/src/agent/configuration_parser/CMakeLists.txt
+++ b/src/agent/configuration_parser/CMakeLists.txt
@@ -1,0 +1,42 @@
+cmake_minimum_required(VERSION 3.18)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_BUILD_TYPE RelWithDebInfo)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
+
+set(CMAKE_TOOLCHAIN_FILE "${CMAKE_SOURCE_DIR}/../../vcpkg/scripts/buildsystems/vcpkg.cmake")
+set(VCPKG_MANIFEST_DIR ${CMAKE_SOURCE_DIR}/../../)
+
+find_package(toml11 CONFIG REQUIRED)
+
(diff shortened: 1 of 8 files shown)


src/agent/configuration_parser/include/toml11.hpp
TomasTurina, line 1: Why is this file here?

src/agent/configuration_parser/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+set(CMAKE_BUILD_TYPE RelWithDebInfo)
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g")
+
+set(CMAKE_TOOLCHAIN_FILE "${CMAKE_SOURCE_DIR}/../../vcpkg/scripts/buildsystems/vcpkg.cmake")
+set(VCPKG_MANIFEST_DIR ${CMAKE_SOURCE_DIR}/../../)
+
+project(ConfigurationParser)
+message(STATUS "******* Project ConfigurationParser *********")
+
+#########################################################################
```
TomasTurina, line 14: Why is this necessary?

src/agent/configuration_parser/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
+cmake_minimum_required(VERSION 3.18)
+
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_BUILD_TYPE Release)
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -g")
```
jr0me, line 6: We should start using `RelWithDebInfo`.

Also, to my understanding `CMAKE_CXX_FLAGS_DEBUG` only sets flags for `Debug` build types, you can use instead `CMAKE_CXX_FLAGS`.

src/agent/configuration_parser/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+
+#########################################################################
+# Dependencies
+#########################################################################
+
+# Enable testing
+enable_testing()
+
+include_directories(include)
+include_directories(${CMAKE_SOURCE_DIR}/configuration_parser)
```
jr0me: I don't think we should include this folder. Already from line 38 the headers for the configuration parser should be visible to the rest of the project.

src/agent/configuration_parser/include/configuration_github.hpp
```diff
@@ -0,0 +1,75 @@
...
+    const bool DEFAULT_ONLY_FEATURE_EVENTS = true;
+    const std::string DEFAULT_EVENT_TYPE = "all";
+
+    class GithubConfigApiAuth
+    {
+        private:
+            std::string orgName;
+            std::string apiToken;
+        public:
+            GithubConfigApiAuth(){};
```
jr0me: There's no no need for a semicolon after the closing } of a function definition or a control structure.

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -16,74 +16,13 @@ namespace configuration
         toml::value tbl;
         configurationGithub::ConfigurationGithubModule configGithub;
 
-        void LoadGithubConfig()
```
jr0me: Since we are separating header and source file in this commit, could we do it in the first commit these files are presented? This will lead to a cleaner git history.

src/agent/configuration_parser/include/configuration_github.hpp
jr0me, line 1: Since this file is not needed and was introduced by this branch in the first place, maybe we can avoid adding it altogether.

jr0me: (Same goes for the rest of the code being removed)

src/agent/configuration_parser/CMakeLists.txt
```diff
@@ -28,18 +32,27 @@ MACRO_ENSURE_OUT_OF_SOURCE_BUILD(
...
-
 include_directories(include)
 include_directories(${CMAKE_SOURCE_DIR})
 
-# Add file to src code
-add_executable(ConfigurationParser_test
-    tests/configuration_parser_test.cpp
-    src/configuration_parser.cpp
-)
+find_package(GTest CONFIG REQUIRED)
```
jr0me, line 38: This package is only needed when building tests, it should be moved inside the if block below.

src/agent/configuration_parser/tests/configuration_parser_test.cpp
```diff
@@ -1,54 +1,144 @@
...
-    {
-        auto agent_map = parser->GetConfig<std::map<std::string, std::string>>("bad_map");
-    }
-    catch(const std::exception& e)
-    {
-        std::cout << "Exception caught" << '\n';
-    }
-
-    return 0;
+using namespace configuration;
```
jr0me, line 6: It would be better to introduce the tests from the beginning instead of creating the `main()` function.

src/agent/configuration_parser/include/toml11.hpp
jr0me: It's been now removed, now toml11 is picked up using Vcpkg.

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include <toml.hpp>
+
+#include <iostream>
+
+namespace configuration
+{
+    class ConfigurationParser
+    {
+    private:
+        std::string configFileName = "wazuh.conf";
```
TomasTurina, line 12: This needs to be modified once the directory structure is defined.

TomasTurina: Create a define for `wazuh.conf`.

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <toml.hpp>
+
+#include <iostream>
+
+namespace configuration
+{
+    class ConfigurationParser
```
TomasTurina, line 9: I think this should be instantiated in the Agent class.

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -0,0 +1,35 @@
...
+        auto GetConfig(Ks... ks)
+        {
+            try
+            {
+                auto config = toml::find<T>(tbl, ks...);
+                return config;
+            }
+            catch (const std::exception& e)
+            {
+                std::cout << "The requested value could not be obtained: " << e.what() << std::endl;
```
TomasTurina, line 30: This should be replaced by a logger once defined.

jr0me: Should we create a story to create such logger?

src/agent/configuration_parser/include/configuration_parser.hpp
jr0me: Done

src/agent/configuration_parser/include/configuration_parser.hpp
jr0me: I created a constant hidden in the source file for the configuration parser. Since it was not parameterized and used only once in the default constructor there's no need to store it as a class instance member.

TomasTurina (approved):

Issue #14, opened 2024-06-25
Develop the new client


vikman90:
**Parent issue:**
- https://github.com/wazuh/wazuh/issues/22677

## Description

Following the completion of the spike on agent-manager communication, the next step is to implement a functional client for this new protocol. This client will establish a fully functional communication and handshake with the server, transmit queue information, and leave the interface open for the addition of other modules.

## Functional requirements

1. Connect to the server.
2. Register and obtain a UUID.
3. Authenticate and acquire a token.
4. Send data using that token. Let such data persist in the agent until it receives a `200` code from the server.
  a. Stateful.
  b. Stateless.
5. Query commands that the server needs to send to the agent.
6. **Resilience:** reconnect to the server if the connection is lost.
7. **Multitasking:** support multiple concurrent connections.
8. Read options from a configuration file.
9. Work alongside the queue system.

## Implementation restrictions

1. We will adhere to the restrictions declared in issue #1.
2. The configuration will be in **TOML** format.
3. The configuration will be parsed in a separate component, such as `ConfigParser`.
4. Support for multiple communications will preferably be implemented without multithreading, using coroutines or similar.
5. If possible, use C++20.

## Plan

1. Start from the PoC developed in issue #1.
2. Agree on the **queue** interface: #2.
3. Agree on the **command manager** interface: #4.
4. Investigate if all target platforms support C++20 with GCC and Clang.
5. Design and implement a task manager #24.
6. Design and implement a communicator module #25.
7. Design and implement a configuration parser #26.
8. Integrate with the Queue component #57.
9. Address technical debt #64.
10. If necessary, implement a "Dummy" module for testing purposes.

vikman90:
## C++20 Concepts vs inheritance

I've conducted a trial to model a module pool in C++, aiming to create a class capable of managing references to all modules, starting them, and facilitating communication. 

To achieve this, I explored two approaches:
- **Inheritance**: Introducing an `IRunnable` interface where modules inherit and implement it.
- **Concepts**: Utilizing a `Runnable` concept, with the pool restricting parameters to this concept.

Initially, conceptual constraints seemed preferable over inheritance to ensure each module remains independent of the "Module" definition. However, I noted that using a `std::vector<std::any>` container — akin to the classic `void*[]` in C — introduces excessive generality. This approach necessitates runtime type inference for each module access operation, thereby:
- Missing compile-time error detection.
[...]

PR #89, opened 2024-08-18
Enhancement/14 add key to register and authentication

sdvendramini:
Related issue

- https://github.com/wazuh/wazuh-agent/issues/14

This PR aims to correct the necessary parameters to complete the registration process and test this function correctly.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -30,9 +30,9 @@ runs:
         shell: bash
 
       - name: Run Tests
-        continue-on-error: true
+        continue-on-error: false
         run: |
           set +e
           cd build
-          ctest || echo "Tests failed, but continuing"
+          ctest
         shell: bash
--- a/src/agent/include/register.hpp
+++ b/src/agent/include/register.hpp
@@ -1,5 +1,9 @@
 #pragma once
 
+#include <http_client.hpp>
+
+#include <boost/beast/http.hpp>
+#include <functional>
 #include <optional>
 #include <string>
 
@@ -11,5 +15,17 @@ namespace registration
         std::string password;
     };
 
-    bool RegisterAgent(const UserCredentials& userCredentials);
+    using RegisterFunctionType = boost::beast::http::status(const std::string& host,
+                                                            const std::string& port,
+                                                            const std::string& token,
+                                                            const std::string& uuid,
+                                                            const std::string& key,
+                                                            const std::optional<std::string>& name);
+
+    RegisterFunctionType SendRegistrationRequest;
+
+    bool RegisterAgent(const UserCredentials& userCredentials,
+                       std::function<http_client::AuthenticateFunctionType> AuthenticateFunction,
+                       std::function<RegisterFunctionType> RegisterFunction);
+
 } // namespace registration
--- a/src/agent/tests/register_test.cpp
+++ b/src/agent/tests/register_test.cpp
@@ -1,12 +1,68 @@
 #include <register.hpp>
 
+#include <agent_info.hpp>
+
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
-TEST(RegistrationTest, RegistrationTest)
+namespace beast = boost::beast;
+namespace http = beast::http;
+using ::testing::_;
+using ::testing::MockFunction;
+using ::testing::Return;
+
+class RegisterTest : public ::testing::Test
 {
-    const registration::UserCredentials userCredentials {"user", "123456"};
-    const bool res = registration::RegisterAgent(userCredentials);
+protected:
+    std::unique_ptr<AgentInfo> agent;
+
+    MockFunction<http_client::AuthenticateFunctionType> MockAuthenticateFunction;
+    MockFunction<registration::RegisterFunctionType> MockRegistrationFunction;
+
+    void SetUp() override
+    {
+        agent = std::make_unique<AgentInfo>("agent_name", "agent_key", "fbdea7a6-0e0d-4d2c-8af3-9330ea8799a3");
+    }
+};
+
+TEST_F(RegisterTest, RegistrationTest)
+{
+    const registration::UserCredentials userCredentials {"user", "password"};
+
+    const std::optional<std::string> name = agent->GetName();
+
+    EXPECT_CALL(MockAuthenticateFunction, Call("localhost", "55000", "user", "password")).WillOnce(Return("token"));
+    EXPECT_CALL(MockRegistrationFunction, Call("localhost", "55000", "token", agent->GetUUID(), agent->GetKey(), name))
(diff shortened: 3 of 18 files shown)


src/agent/include/register.hpp
```diff
@@ -1,8 +1,16 @@
 #pragma once
 
+#include <http_client.hpp>
```
jr0me, line 3: Is this include necessary?

src/agent/include/register.hpp
```diff
@@ -1,8 +1,16 @@
...
+
+#include <boost/beast/core.hpp>
```
jr0me, line 5: I'm not sure that we need both core and http in the header file.

src/agent/include/register.hpp
```diff
@@ -1,8 +1,16 @@
...
+#include <boost/beast/http.hpp>
+#include <functional>
 #include <optional>
 #include <string>
 
+namespace beast = boost::beast;
```
jr0me, line 11: Beast doesn't seem necessary, but I think that it is preferable to not introduce these namespace declarations on header files, any other file including this header will "inherit" them, polluting the global namespace.

src/agent/include/register.hpp
```diff
@@ -11,5 +19,21 @@ namespace registration
         std::string password;
     };
 
-    bool RegisterAgent(const UserCredentials& userCredentials);
+    bool RegisterAgent(
+        const UserCredentials& userCredentials,
+        std::function<std::optional<std::string>(
+            const std::string&, const std::string&, const std::string&, const std::string&)> AuthenticateFunction,
```
jr0me, line 25: Though parameter names are not needed here, given that amount of arguments, we should better include them so readers of this function signature know what it's about. It's also weird to have RegistrationFunction with each argument named, but AuthenticateFunction with none.

Furthermore, since the `RegisterAgent` declaration is a bit too verbose we could try to use type alises or decltype to make it more legible.

```c++

    using RegistrationFunctionType = http::status(const std::string& host,
                                                  const std::string& port,
                                                  const std::string& token,
                                                  const std::string& uuid,
                                                  const std::string& key,
                                                  const std::optional<std::string>& name);

    RegistrationFunctionType SendRegistrationRequest;

    bool RegisterAgent(
        const UserCredentials& userCredentials,
        std::function<std::optional<std::string>(
            const std::string&, const std::string&, const std::string&, const std::string&)> AuthenticateFunction,
        std::function<RegistrationFunctionType> RegistrationFunction);
```

The same could be done for AuthenticateFunction, but the function type/signature, should be declared in http_client.

jr0me: Also, nitpicking a bit here, but could we make the names "symmetrical"? Ie: AuthenticateFunction and RegisterFunction, or AuthenticationFunction and RegistrationFunction.

src/agent/tests/register_test.cpp
```diff
@@ -1,12 +1,57 @@
...
 
-TEST(RegistrationTest, RegistrationTest)
+using ::testing::_;
+using ::testing::MockFunction;
+using ::testing::Return;
+
+class RegisterTest : public ::testing::Test
+{
+protected:
+    const AgentInfo* agent;
```
jr0me, line 15: Please use `unique_ptr` here.

src/agent/tests/register_test.cpp
```diff
@@ -1,12 +1,57 @@
...
+        MockAuthenticateFunction;
+    MockFunction<http::status(const std::string& host,
+                              const std::string& port,
+                              const std::string& token,
+                              const std::string& uuid,
+                              const std::string& key,
+                              const std::optional<std::string>& name)>
+        MockRegistrationFunction;
+
+    ~RegisterTest()
```
jr0me, line 28: No need for a custom destructor if using smart ptr.

src/agent/tests/register_test.cpp
```diff
@@ -1,12 +1,57 @@
...
+    {
+        delete agent;
+    }
+    void SetUp()
+    {
+        agent = new AgentInfo("agent_name", "agent_key", "fbdea7a6-0e0d-4d2c-8af3-9330ea8799a3");
```
jr0me, line 34: Use make_unique

src/agent/tests/register_test.cpp
```diff
@@ -1,12 +1,57 @@
...
+    }
+    void TearDown()
+    {
+        delete agent;
```
jr0me, line 38: Better to use reset with unique ptr https://en.cppreference.com/w/cpp/memory/unique_ptr/reset

src/agent/tests/register_test.cpp
```diff
@@ -1,12 +1,57 @@
...
+        agent = nullptr;
+    }
+};
+
+TEST_F(RegisterTest, RegistrationTest)
```
jr0me, line 43: Now that we can mock the functions used, we could write tests that test: when SendRegistrationRequest or AuthenticateFunction fail.

.github/actions/compile/action.yml
```diff
@@ -30,7 +30,7 @@ runs:
         shell: bash
 
       - name: Run Tests
-        continue-on-error: true
+        continue-on-error: false
```
jr0me, line 33: It's also necessary to remove the `|| echo "Tests failed, but continuing` from line 37.

jr0me (changes requested):
I noted a few things here and there but looking good!

src/agent/tests/register_test.cpp
jr0me: Actually, since we are creating a new agent on `SetUp`, there's no need to call `reset`. It will happen automatically on assignment.

src/agent/include/register.hpp
sdvendramini: After a few changes, now it's needed to use `http_client::AuthenticateFunctionType`

src/agent/include/register.hpp
sdvendramini: Fixed!

src/agent/include/register.hpp
sdvendramini: Fixed!

src/agent/include/register.hpp
sdvendramini: Done!

src/agent/tests/register_test.cpp
sdvendramini: Done!

src/agent/tests/register_test.cpp
sdvendramini: Done!

src/agent/tests/register_test.cpp
sdvendramini: Done!

TomasTurina (approved):

PR #98, opened 2024-08-21
Fix includes location and names

TomasTurina:
## Description

This PR includes the following fixes:

- Fix the location of the `.hpp` files. Those that are not public interfaces were moved to the corresponding `src` folder.
- Rename `.h` files to `.hpp` to keep a standardized naming.
- In addition, the usage of a string to identify a persistence was change by an `enum`.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X


src/agent/multitype_queue/src/persistence.hpp
jr0me, line 1: `multitype_queue.hpp` needs this header. So it needs to be in the `include` public folder.

src/agent/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ set(SOURCES
 )
 
 add_library(Agent ${SOURCES})
-target_include_directories(Agent PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(Agent PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
jr0me, line 30: If we go down this road, this added directory here should be `PRIVATE`, otherwise it is still visible to everyone that includes this target.

src/agent/src/task_manager.hpp
jr0me, line 1: This file is included in the `agent.hpp`, it needs to be public.

src/agent/src/itask_manager.hpp
jr0me, line 1: This file also needs to be public, since it is included in the public header `task_manager.hpp`

src/agent/src/signal_handler.hpp
jr0me, line 1: Also included in agent.hpp, needs to be public.

src/agent/multitype_queue/src/persistence_factory.hpp
```diff
@@ -7,8 +7,13 @@
 #include <string>
 #include <vector>
 
+enum PersistenceType
```
jr0me, line 10: Can this be an `enum class`?

Also, can it be nested inside the public part of `PersistenceFactory` ?

src/agent/multitype_queue/src/persistence_factory.hpp
```diff
@@ -7,8 +7,13 @@
 #include <string>
```
jr0me, line 7: The `string` header is no longer needed here, but could be needed in the source file...

jr0me (changes requested):
Personally I don't like mixing header files and source files in a folder called `src`. Maybe we can have a folder inside `src` called `internal`, or maybe the `internal` folder is at the same level of `include` and `src`. I wonder if there are other common patterns we could copy.

When you include directories publicly in a target, every one that uses it will see those includes, so it defeats the purpose of moving any private file to that directory. Beware to include them as `PRIVATE`.

Note that there's also the possibility of creating `INTERFACE` targets, where we could only expose the public headers. Consumers of it, and the target with the implementation would link to it, further encapsulating the private target since consumers would only link to the interface targets. It's an option, may be overkill.

src/agent/communicator/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ find_package(nlohmann_json CONFIG REQUIRED)
 find_path(JWT_CPP_INCLUDE_DIRS "jwt-cpp/base.h")
 
 add_library(Communicator src/communicator.cpp src/http_client.cpp)
-target_include_directories(Communicator PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include ${JWT_CPP_INCLUDE_DIRS})
+target_include_directories(Communicator PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include ${JWT_CPP_INCLUDE_DIRS} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
jr0me, line 17: If no files from `Communicator` were moved, we can undo this.

src/agent/configuration_parser/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ set_common_settings()
 find_package(toml11 CONFIG REQUIRED)
 
 add_library(ConfigurationParser src/configuration_parser.cpp)
-target_include_directories(ConfigurationParser PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(ConfigurationParser PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
```
jr0me, line 14: Same, it seems that there's no reason to add the `src` folder as a target include.

src/agent/communicator/tests/CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 find_package(GTest CONFIG REQUIRED)
 
 add_executable(communicator_test communicator_test.cpp)
+target_include_directories(communicator_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include ${CMAKE_CURRENT_SOURCE_DIR}/../src)
```
jr0me, line 4: These changes belong to the previous commit.

src/agent/configuration_parser/tests/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 find_package(GTest CONFIG REQUIRED)
 
 add_executable(ConfigurationParser_test configuration_parser_test.cpp)
+target_include_directories(ConfigurationParser_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include ${CMAKE_CURRENT_SOURCE_DIR}/../src)
```
jr0me, line 4: Same as above.

src/agent/multitype_queue/src/persistence_factory.cpp
```diff
@@ -3,10 +3,10 @@
 
 #include <stdexcept>
```
jr0me, line 4: Since we removed the `string` header from the hpp file, we need to add it here in the source file.

src/agent/multitype_queue/tests/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 find_package(GTest REQUIRED)
 
 add_executable(test_MultiTypeQueue multitype_queue_test.cpp)
+target_include_directories(test_MultiTypeQueue PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include ${CMAKE_CURRENT_SOURCE_DIR}/../src)
```
jr0me, line 4: Should go in previous commit.

src/agent/tests/CMakeLists.txt
```diff
@@ -1,21 +1,26 @@
 find_package(GTest CONFIG REQUIRED)
 
 add_executable(agent_test agent_test.cpp)
+target_include_directories(agent_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include ${CMAKE_CURRENT_SOURCE_DIR}/../src)
```
jr0me, line 4: Also should go in previous commit.

jr0me (changes requested):
Almost there, but much better! 🚀

jr0me (approved):
💯

Issue #83, opened 2024-08-14
Develop the Agent Command Handler


lchico:
**Parent issue:**
- https://github.com/wazuh/wazuh/issues/22887

## Description
Based on issue #4, we need to implement a command manager that executes the commands placed in the queue and provides feedback.

The following diagram is taken from [#4](https://github.com/wazuh/wazuh-agent/issues/4#issuecomment-2180316838)

```mermaid
sequenceDiagram
    participant Server
    participant AgentCommsAPIClient
    participant CommandManager
    participant CommandStore
    participant Executor
    participant Feedback

    AgentCommsAPIClient->>Server: getConnection()
    AgentCommsAPIClient->>Server: pollCommands()
    Server->>AgentCommsAPIClient: sendCommand()
    AgentCommsAPIClient->>CommandManager: addCommands()
    CommandManager->>CommandStore: storeCommand()
    CommandManager->>Executor: executeCommand()
    Executor->>CommandStore: getNextCommand()
    Executor->>Executor: execute()
    Executor->>Feedback: generateFeedback()
    Executor->>CommandManager: reportFeedback()
    CommandManager->>CommandStore: markCompleted()
    CommandStore->>CommandStore: deleteCompletedCommands()
    CommandManager->>AgentCommsAPIClient: reportFeedback()
    AgentCommsAPIClient->>Server: reportFeedback()
```

## Functional requirements
1. Commands persist on the agent: upon execution, they are marked as completed and reported to the manager.
2. Each command generates a result indicating completion status and outcome (success or error) with enough clarity for the user to take appropriate action.
3. Completed commands are deleted from the agent once the manager confirms receipt of the result.
4. All commands must be executed in the order of creation.

## Implementation restrictions

1. **Language:**
    We want to use C++ to leverage its performance and system-level capabilities.
2. **Modular integration:**
    This module should be implemented as a library to ensure modularity and reusability, allowing it to be integrated seamlessly with other modules.
3. **Process tied:**
    This module will be tied to the main process of the agent, ensuring it operates as a core component of the agent's functionality.

aritosteles:
# Update
- **(2024/08/14)**: Understanding issue requirements and getting acquainted with status of related issues.
- - Created work branch and added basic module structure.
- - Added placeholder coroutine to get commands from queue.
- - At the moment @LucioDonda is working on the Queue module and @jr0me is working on the integration with the existing modules.

LucioDonda:
[...]

PR #101, opened 2024-08-22
Enhancement/83 persistence

aritosteles:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/83|

## Description
This PR aims to turn existing SQLiteStorage class, which is currently coupled with MultiTypeQueue class, into a generic sqlite wrapper that can be used by any module that requires SQLite database functionality.

## Tests

  - [ ] Linux


src/agent/multitype_queue/src/sqlitestorage.cpp
```diff
@@ -6,7 +6,38 @@
 #include <exception>
 #include <iostream>
 
-SQLiteStorage::SQLiteStorage(const std::string& dbName, const std::vector<std::string> tableNames)
+std::string SQLiteStorage::GenerateQuery(const std::string& formatString, const std::vector<std::string>& params)
```
jr0me, line 9: Isn't it possible to use `fmt` here?

src/agent/multitype_queue/include/ipersistence.hpp
TomasTurina, line 1: I think having `queueName` and `moduleName` as part of the function parameters doesn't make this interface generic.

src/agent/multitype_queue/include/multitypequeue_queries.hpp
TomasTurina, line 1: I think this file can be moved to src folder.

src/agent/multitype_queue/src/sqlitestorage.cpp
TomasTurina: Can't we make SQL queries `constexpr`?
https://github.com/wazuh/wazuh-agent/pull/101/files#diff-80d1e3e3f3442f1af6f1658d4ccc5cd2a7815de0e850484ede01b034a778c43dR5

src/agent/multitype_queue/include/sqlitestorage.hpp
TomasTurina, line 1: Same, many function have parameters that are not generic.

jr0me: Maybe we should discuss _how_ generic we want this class to be. The entire interface assumes that the table has only two columns (three if you count the id) if I'm not wrong... do we know how many columns we will need outside the MultiTypeQueue case? The range from somewhat specialized to fully generic is huge.

aritosteles: Right, it also assumes only string types. Making it full generic would imply allowing all (or most usual) types.

src/agent/multitype_queue/src/sqlitestorage.cpp
jr0me: Found this workaround, it may work https://godbolt.org/z/ebaKxv9Gh

src/agent/multitype_queue/include/sqlitestorage.hpp
TomasTurina: I mean if we parametrize the query we can also parametrize the list of filters and/or values, i.e using a vector of any. The caller should know what filters/values pass to the function depending on the query to execute. What do you think?

jr0me: I agree, we could construct the table and the queries from a list of "columns".

src/agent/multitype_queue/src/sqlitestorage.cpp
aritosteles: It works. Done.

Issue #66, opened 2024-08-05
Design a Pipeline for Running Smoke Tests with Specific Components


TomasTurina:
## Description

After [analysis](https://github.com/wazuh/wazuh-agent/issues/38) of suitable tools and systems to manage dependencies in Wazuh, the decision was made to use `vcpkg`. Since it has already been [adopted](https://github.com/wazuh/wazuh-agent/issues/58) in agent repositories, it is now necessary to create a pipeline to run smoke tests with specific components using the `vcpkg` dependencies.

## Tasks

1. **Define the stages of the smoke testing pipeline, including dependency retrieval, component compilation, and testing**:
    - Use Github Actions to do this.
    - For testing consider using the API to connect and check that the expected modules are available. 

2. **Link your smoke testing pipeline with your dependency build pipeline to ensure compatibility**:
    - Dependency build pipeline: https://github.com/wazuh/wazuh-agent/issues/65

3. **Ensure that the pipeline supports all required platforms**:
    - Platforms: Windows, Linux, and macOS.

4. **Documentation**:
    - Document the steps required to follow this procedure.

## Acceptance Criteria

- A working pipeline for each supported OS capable of building and testing specific components with `vcpkg` dependencies.
  - There should be a workflow for each supported OS on Github Actions.
  - It should work for the agent components.
- A documented guide of the steps required to run smoke tests with specific components.

## Related Issues

- **Parent:** https://github.com/wazuh/wazuh/issues/16882

lchico:
# Update
- I did some research on how to implement it and started doing a proof of concept:

<details>
<summary>Module agent_info</summary>
<mark>Note:</mark> I created the <strong>vcpkg.json</strong>  file with the dependencies required by the module and updated the <strong>CMakeLists.txt</strong> file.

Changes required:

```diff
git diff
diff --git a/src/agent/agent_info/vcpkg.json b/src/agent/agent_info/vcpkg.json
new file mode 100644
index 000000000..a514a305c
--- /dev/null
+++ b/src/agent/agent_info/vcpkg.json
@@ -0,0 +1,8 @@
+{
+    "name": "wazuh-agent-mvp",
+    "version": "5.0.0",
+    "dependencies": [
+      "boost-uuid",
+      "sqlitecpp"
+    ]
+}

diff --git a/src/agent/agent_info/CMakeLists.txt b/src/agent/agent_info/CMakeLists.txt
index 3d8d1b371..f92597dc8 100644
[...]

PR #116, opened 2024-09-05
Adds a smoke test pipeline for specific component

lchico:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/66|

## Description
As a proof of concept, I created the following pipeline.

**Output:**  The goal is to use the API to validate the status of the modules. At present, we only confirm that the binary is successfully generated and executable.

**Notes:**
1. For the POC, we used Linux.
2. For testing, since the API is not ready, we only tested that the built binary can be executed.

```mermaid
graph TD
    A[Start CI/CD Pipeline] --> B[Select Component to Build]
    B --> C1[Agent]
    C1 --> F1[Select modules: <br> Ex. FIM, syscollector]
    C1 --> D[Install Required Dependencies]
    D --> E[Build]
    E --> K{Test Modules <br> API status request}
    F1 --> K
    K --> |Yes| L[Deploy Main Program]
    K --> |No| M[Notify Failure and Abort]
```

## Tests
Command to run the workflow:
```
gh workflow run build_and_test_component.yml -r enhancement/66-design-a-pipeline-for-running-smoke-tests-with-specific-components
```

The workflow has been tested: 
https://github.com/wazuh/wazuh-agent/actions/runs/10962476015/job/30441808451 :heavy_check_mark:

### DIFF
--- a/.github/workflows/build_and_test_component.yml
+++ b/.github/workflows/build_and_test_component.yml
@@ -1,0 +1,56 @@
+run-name: Build and Test Component - Build ${{ inputs.component }}
+name:  Build and Test Component
+
+on:
+  workflow_dispatch:
+    inputs:
+      source_reference:
+        description: |
+          Branch from wazuh/wazuh-agent repository to use.
+        required: true
+      component:
+        description: 'Default agent'
+        required: true
+        type: choice
+          - agent
+        default: agent
+
+  workflow_call:
+    inputs:
+      source_reference:
+        type: string
+        required: true
+      components:
+        type: string
+        required: true
+
+jobs:
+  Build-component-and-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    name: Build ${{ inputs.components }}
+
+    steps:
+      - name: Checkout wazuh/wazuh-agent repository
+        uses: actions/checkout@v4
+        with:
+          repository: wazuh/wazuh-agent
+          ref: ${{ inputs.source_reference }}
+
+      - name: Initializing Submodules
+        run: |
+          git submodule update --init --recursive
+          cd src/vcpkg && ./bootstrap-vcpkg.sh
+
+      - name: Build component
+        if: ${{ inputs.component == 'agent' }}
+        run: |
+          mkdir src/build
+          cd src/build
+          cmake .. -DBUILD_TESTS=ON
+          make
+
+      - name: Run the tests
+        run: |
+          cd src/build && ./wazuh-agent &> status.log & sleep 3; pkill wazuh-agent
+          cat src/build/status.log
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test_component.yml
```diff
@@ -0,0 +1,56 @@
...
+        type: choice
+          - agent
+        default: agent
+
+  workflow_call:
+    inputs:
+      source_reference:
+        type: string
+        required: true
+      components:
```
mjcr99, line 23: This input variable has an `s` in the end, shouldn't it be the same as the one used in the `workflow_dispatch` part?

.github/workflows/build_and_test_component.yml
```diff
@@ -0,0 +1,56 @@
...
+    name: Build ${{ inputs.components }}
+
+    steps:
+      - name: Checkout wazuh/wazuh-agent repository
+        uses: actions/checkout@v4
+        with:
+          repository: wazuh/wazuh-agent
+          ref: ${{ inputs.source_reference }}
+
+      - name: Initializing Submodules
```
mjcr99, line 40: This step could be removed since the previous one can be set to fetch the `wazuh-agent` repository with its submodules given a certain `ref`. An example can be seen here:
* https://github.com/wazuh/wazuh-agent/blob/524768a0ca71c1ecdd52d742f3923842a672b118/.github/workflows/compile_and_run_tests.yml#L20

.github/workflows/build_and_test_component.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          git submodule update --init --recursive
+          cd src/vcpkg && ./bootstrap-vcpkg.sh
```
mjcr99, line 43: This line may not be necessary since the bootstrap is automatically done in the next step when `cmake .. -DBUILD_TESTS=ON` is run.

mjcr99:
@lchico regarding Binary Caching step:
Addig this step the same as it's in the shared file, before running the compilation should enable the binary caching:
* https://github.com/wazuh/wazuh-agent/blob/d0f7ed5c14f74a96a480308eea310a5e8d2ba65b/.github/workflows/compile_and_run_tests.yml#L43

Also have a look at this step, which adds some dependencies that are not currently covered in your workflow:
* https://github.com/wazuh/wazuh-agent/blob/d0f7ed5c14f74a96a480308eea310a5e8d2ba65b/.github/workflows/compile_and_run_tests.yml#L29

mjcr99 (approved):
Added 1b02c9cea1423012782f3757cdd3376c1911b059 addressing minnor fixes. LGTM, GJ @lchico!

MarcelKemp (approved):

Issue #65, opened 2024-08-05
Design a Pipeline for Uploading New Dependencies


TomasTurina:
## Description

After [analysis](https://github.com/wazuh/wazuh-agent/issues/38) of suitable tools and systems to manage dependencies in Wazuh, the decision was made to use `vcpkg`. Since it has already been [adopted](https://github.com/wazuh/wazuh-agent/issues/58) in both agent and manager repositories, it is now necessary to create a pipeline to upload new dependencies for `vcpkg` when needed.

## Tasks

1. **Define the pipeline stages, including dependency validation, packaging, and loading**:
    - Consider all dependencies, including those not in the `vcpkg` repository.
    - Use Github Actions to do this.

2. **Ensure that the pipeline supports all required platforms**:
    - Platforms: Windows, Linux, and macOS.

3. **Documentation**:
    - Document the steps required to follow this procedure.

## Acceptance Criteria

- A working pipeline for each supported OS capable of validating, packaging, and loading dependencies to be used by `vcpkg`.
  - There should be a workflow for each supported OS in Github Actions.
  - It should work for both agent and manager dependencies.
- A documented guide of the steps required to load new dependencies into `vpckg` using the designed pipeline.

## Related Issues

- **Parent:** https://github.com/wazuh/wazuh/issues/16882

mjcr99:
## Update 
**(30/08/2024)** Researching related issues and documentation about `vcpk` capabilities. Understanding issue purpose. Found some doubts to clarify with the team:

* Where is intended to store the dependencies? I have been reading about https://github.com/marketplace/actions/cache, but it seems to be oriented to store dependencies used by workflows, not by external code, as would be our case.
* Researching [Tutorial: Set up a vcpkg binary cache using GitHub Packages in a GitHub Actions workflow](https://learn.microsoft.com/en-us/vcpkg/consume/binary-caching-github-packages?pivots=linux-runner) it seems it focused on Nuget packages which are `.NET` packages.
* It's also needed to fully understand a case where the uploaded dependencies are pulled to build an agent, to understand completely hoy these dependencies should be stored.

mjcr99:
## Design Proposal

After researching the possibilities for VCPKG and running a POC the following design is proposed:

[...]

PR #117, opened 2024-09-06
Add upload dependencies workflows

mjcr99:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/65|

## Description

Hi team, this PR enables the following features:
* `vcpkg` binary caching: Using the GitHub packages storage in combination with NuGet, we can store the precompiled dependencies we use in the building procedure. This reduces the workflow's building time, resulting in a faster execution. This binary caching has been set up in all the workflows that currently execute a building procedure.

* Port-overlay for `procps`, `libdb`, and `librpm`: These three dependencies are not natively available in `vcpkg`. To use them 3, port-overlays have been developed to compile and manage these libraries directly from `vcpkg`, allowing us to manage al the dependencies directly with `vcpkg`. For the specific case of `librpm` the following dependencies has been added to the workflows as they are needed to build this library:

```console
sudo apt-get install -y autopoint libtool zlib1g-dev libgcrypt20-dev libmagic-dev libpopt-dev libmagic-dev libsqlite3-dev liblua5.4-dev gettext libarchive-dev
```


MarcelKemp (approved):
LGTM!

vikman90 (approved):

PR #151, opened 2024-09-12
Windows compilation with MSVC requirements

ncvicchi:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/17|

## Description

Wazuh's agent legacy code was expected to be compiled using MinGW, which allowed that almost the same code for Linux to be compiled for Windows.
Due to several disadvantages of that approach, moving to MSVC was determined, but this impacted the code with several incompatibilities.
This PR aims to generate a compilable code (but not a working one) for the syscollector to inventory migration branch. It actually solves several compiler and platform incompatibilities, but since it mocks almost all required functions not available on Windows and/or MSVC, it is not yet ready to be tested.

The following section will present an analysis, of the incompatibilities found, the approach used to solve them, and hazards that might require further study to overcome/prevent.

## Local compiling for Windows
At this state of development, these are the requirements and steps to successfully compile the code:

- Windows Server 2022 was used, although probably any Windows will be ok
- Microsoft Visual Studio 2022 with all options enabled (probably most of them unnecessary, should be tested the minimum requirements).
- CMake 3.30.3 was installed and used
- Make was installed through chocolatey (on Powershell):

```console

Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))

choco insall make
```
- Git for Windows was used (not mandatory)

Once cloned the repository and moved to the branch, from the `src` directory:

```console

git submodule update --init --recursive
mkdir build
cd build
cmake ..  -DBUILD_TESTS=1 -G "Visual Studio 17 2022" -A x64
cmake --build .

```

To verify the generated binary:
```console

.\Debug\wazuh-agent.exe --register

```

### Conditional behaviour
Among legacy and new code, to decide whether some code/directive must be used, conditional preprocessing is used. But these conditions are based on different approaches, and a decision to normalize it should be taken.

Should these decisions be made on the platform (WIN32, Windows, Linux, Darwin) or compiler (MSVC, GCC, clang, \_\_GNUC\_\_, \_\_clang\_\_)? Common criteria should be established and kept throughout all the code, both at cmakelist and source codes.

### Cmakelists

All cmakelist files expect GCC as the target compiler. MSVC doesn't support GCC flags, thus failing compilation. Almost all GCC flags are not available at MSVC, thus losing their behavior selection. Some examples:
Particularly, `/EHsc` had to be added in several cmakelist to avoid errors on compilation.

GCC:  `"-g -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Woverloaded-virtual -Wunused -Wcast-align -Wformat=2 -std=c++14 -pthread"`
GCC:    `"-fsanitize=address,leak,undefined"`
MSVC: `"/Zi /W4 /permissive- /std:c++14 /MT /EHsc"`

GCC: `"-Wall -Wextra -Wshadow -Wnon-virtual-dtor -Woverloaded-virtual -Wunused -Wcast-align -Wformat=2 -std=c++14 -pthread"`
MSVC: `"/W4 /permissive- /std:c++14 /MT"`

GCC: `"-Wall -Wextra -Wshadow -Wnon-virtual-dtor -Woverloaded-virtual -Wunused -Wcast-align -Wformat=2 -std=c++14 -pthread"`
clang:  `"-O3"`
GCC: `"-O3 -s"`
GCC & clang: `"-g -fsanitize=address,leak,undefined"´
MSVC: `"/W4 /permissive- /std:c++14 /MT /EHsc"`

GCC: `"-Wall -Wextra -std=c++14 --coverage"`
MSVC: `"/W4 /permissive- /std:c++14 /MT /coverage"`

Some directories had to be added to Windows compilation cmakelist, that were not necessary on Linux or clang.

### Compiler pragmas
Compiler code options had to be conditionally removed for MSVC:

```code
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-function"
#pragma GCC diagnostic ignored "-Wreturn-type"
#pragma GCC diagnostic ignored "-Wcast-function-type"
#pragma GCC diagnostic pop
```

### Ambigous code
`makeUnique.h` source code was completely removed. Although GCC handled it properly, MSVC was not able to disambiguate it, and it was only necessary for versions < C++14, which are unsupported anyway.

### Compiler attributes
There are GCC attributes set to certain functions or parameters that are not available and there is no equivalence in MSVC.

All these replacements were defined in every file that required. It should be decided if some compatibility common file should be used instead.

GCC: `#define ATTR_NONNULL __attribute__((nonnull))`
MSVC: `#define ATTR_NONNULL`

GCC: `#define ATTR_NONNULL_FIRST __attribute__((nonnull(1)))`
MSVC: `#define ATTR_NONNULL_FIRST`

GCC: `#define ATTR_RET_NONNULL __attribute__((__returns_nonnull__))`
MSVC: `#define ATTR_RET_NONNULL __attribute__((__returns_nonnull__))`

GCC: #define ATTR_NONNULL_TWO __attribute__((nonnull(2)))
MSVC: #define ATTR_NONNULL_TWO

GCC: `#define ATTR_UNUSED __attribute__((unused))`
MSVC: `#define ATTR_UNUSED`

GCC: `#define UNREFERENCED_PARAMETER(P)`
MSVC: `#define UNREFERENCED_PARAMETER(P) (P)`

### Array defined on variable values
Some arrays were defined with const variables, which is not supported by MSVC:

Original
```
    const DWORD vsize = 1024;
    TCHAR value[vsize];
    DWORD dwCount = vsize;
```
Changed:
```
    #define VSIZE 1024
    TCHAR value[VSIZE];
    DWORD dwCount = VSIZE;
```

And some arrays were defined on received parameters, as variable-length arrays, which is not supported _by default_ by MSVC:

Original:
```code
static std::string exec(const std::string& cmd, const size_t bufferSize = 128)
    {
        const std::unique_ptr<FILE, FileSmartDeleter> file{popen(cmd.c_str(), "r")};
        char buffer[bufferSize];
```

Updated:
```code
#include <vector>
...
        std::vector<char> buffer(bufferSize);
        std::string result;

        if (file)
        {
            while (fgets(buffer.data(), bufferSize, file.get()))
            {
                result += buffer.data();
            }
        }

```

In C like code:
Original:
```code

 /* Create temporary file */
        char tmp_file[strlen(final_name) + 7];

```

```code

        char *tmp_file = malloc(strlen(final_name) + 7);
        if(!tmp_file){
            merror("Unmerging '%s': could not reserve temporary memory for '%s'", finalpath, files);
            state_ok = 0;
        }
...
free(tmp_file);
```

### Mocked functions and libraries and types
Some functions or event types that are defined in POSIX libraries are not available. To allow compilation these were mocked. Some examples:

WINAPI was not available:
`typedef NETIOAPI_API (WINAPI* ConvertLengthToIpv4Mask_t)(ULONG, PULONG);`
`typedef NETIOAPI_API (WINAPI* GetIfEntry2_t)(PMIB_IF_ROW2);`

Both `struct tm` and `localtime_r` are not available, since time.h is a POSIX library.
`struct tm *localtime_r(const time_t *timer, struct tm *result){}`

ssize_t not available (used in several parameters and return values.
Changed to `int64_t`

File & directories definitions and functions not available:

```

typedef int DIR;
struct dirent {
    unsigned short d_reclen;    /* length of this record */
    unsigned char  d_type;      /* type of file; not supported
                                   by all file system types */
    char           d_name[256]; /* filename */
};
DIR * opendir(DIR *dir){ return NULL; }
int closedir(DIR *dir){ return 0; }
static char *dirname(char *s){ return NULL; }
static struct dirent *readdir(DIR *dir){ return NULL; }
static int64_t ftello64(FILE *x){ return 0; }
static int64_t fseeko64(FILE *x, int64_t pos, int mode){ return 0; }
static int64_t S_ISREG(int64_t flags){ return 0; }
static int64_t S_ISDIR(int64_t flags){ return 0; }
static char * PathFindFileNameA_(char *s){ return NULL; }
static void PathRemoveFileSpec_(char *path){}

```

Not available:
char *strtok_r(char *str, const char *delim, char **saveptr){return NULL; }

### Recursive circular inclusion
Although #ifdef HEADER_NAME (then changed to #pragma once as requested) was present to avoid it, on MSVC a circular inclusion occur which prevented from compiling on MSVC, and GCC handled successfully.
To avoid it, some definitions that were use both in shared.h and another headers were removed from it to include them separately, and a new temporary header was used (cust_types.h).

These were not available at link time:

`PathFindFileNameA`
`PathRemoveFileSpec`

They had to be renamed and mocked to

`PathFindFileNameA_`
`PathRemoveFileSpec_`

### Tests 
AgentTest and inventoryTest had to be disabled since this is still a mocked version

```
if(NOT WIN32)
add_test(NAME AgentTest COMMAND agent_test)
endif()
```

### void * pointers have unknown size


src/modules/inventory/tests/inventoryImp/CMakeLists.txt
```diff
@@ -8,4 +8,6 @@ target_link_libraries(inventoryimp_unit_test PRIVATE
     GTest::gtest_main
     GTest::gmock
     GTest::gmock_main)
+if(NOT WIN32)
 add_test(NAME InventoryImpTest COMMAND inventoryimp_unit_test)
+endif()
```
LucioDonda, line 13: this means that this test is not available under win32? if it's so, shouldn't we directly avoid to add this directory on a directory above?

src/common/cmdHelper/include/cmdHelper.h
```diff
@@ -32,14 +37,14 @@ namespace Utils
     static std::string exec(const std::string& cmd, const size_t bufferSize = 128)
     {
         const std::unique_ptr<FILE, FileSmartDeleter> file{popen(cmd.c_str(), "r")};
-        char buffer[bufferSize];
+        std::vector<char> buffer(bufferSize);
```
LucioDonda, line 40: why using std::vector<char> and not directly std::string ?

src/common/os_regex/src/os_regex_compile.c
```diff
@@ -13,10 +13,15 @@
 #include <stdlib.h>
 #include <ctype.h>
 
-#include <pthread.h>
 #include "shared.h"
 #include "os_regex_internal.h"
 
+#if defined(__GNUC__) || defined(__clang__)
+#include <pthread.h>
+#else
```
LucioDonda, line 21: unnecesary line

LucioDonda (commented):
LGTM but leave some comments

src/modules/inventory/tests/inventoryImp/CMakeLists.txt
ncvicchi: This is temporary. These tests **must** be present and passed. But since this is still with several functions mocked with empty functions, they are not able to run.

ncvicchi: We will add a ToDo commentary to remind to restore these tests once mocked functions are implemented

Issue #152, opened 2024-09-13
Replace `os_regex` with `RE2` in the Agent


cborla:
**Parent issue:**
- [Migrate Syscollector module to new agent](https://github.com/wazuh/wazuh-agent/issues/17)

## Description
Deprecate or remove the non-standard **os_regex** library and replace it with the standard **RE2** library. This will improve compatibility and long-term maintainability of the Inventory module.

## Tasks
- [x] Check if 5.0's agent suits all the conditions to use `RE2`.
- [x] Identify where `os_regex` is used in the codebase.
- [ ] Replace `os_regex` with `RE2`.
- [ ] Refactor the affected code for compatibility.
- [ ] Test the implementation to ensure functionality.

## Testing tasks
- [ ] Ensure all unit tests pass after the replacement.
- [ ] Test compatibility on all supported platforms.

LucioDonda:
## Update 16/09/2024

* Check if 5.0's agent suits all the conditions to use RE2.:
  * based on some analysis it can be used without any problem, from c++14 it works. included in vcpk and basic examples checked.
  * Additional performance check [here](https://github.com/HFTrader/regex-performance?tab=readme-ov-file)
* Identify where os_regex is used in the codebase:
  * **OS_PRegex**: os_string (1) and maybe in the configurations parser.
  * **w_regexec**: version_op (7) in the future if used maybe on remoted_op and wdb_global
  * **w_sql_regex**: Not used since 4.8.0 .
* Starting Replacement generating based UTs to be used for check In the newest version.

LucioDonda:
## Update 17/09/2024

* Starting migration. 
  * working on some issues regarding compilation of C and C++ alongside

LucioDonda:
## Update 18/09/2024

After the team daily I came across that I was pointing to the wrong file. New analysis was made:

1) `int OSRegex_Compile(const char *pattern, OSRegex *reg, int flags);`
Brief: Compile a regular expression to be used later
Usage:
	* expression.c (1)
Tests (wrap):
	* test_expression (4)
	* github unit tests (8)

2) `const char *OSRegex_Execute(const char *str, OSRegex *reg) __attribute__((nonnull(2)));`
Brief: Compares an already compiled OSRegex regular expression with a string
Usage:
	* Internal Only (os_regex.c and os_regex_execute.c)
Tests (wrap):
	* wm_github

[...]

PR #163, opened 2024-09-19
Removal of Dependencies to Libraries Before Migration

LucioDonda:
|Related issue|
|---|
|#152|

## Description

Although the base idea was to replace the regex library with re2 as stated on the base issue, we have proceded with a first removal of any reference of this specially related to inventory which is the parent development.

## Hypothesis

Because we're dealing with a great amount of legacy code that we don't really want to keep using we have decided to first cut any relation with the os_regex library.

## Proposal

* remove any addition of os_regex as library.
* remove any os_regex header inclution.
* comment regex's functions used in libraries that we still need.
* start in inventory module and proceed with any other code black that make use of os_regex

## Tests

  - [x] Linux
  - [ ] Windows
  - [x] MAC OS X


src/common/debug_op/CMakeLists.txt
```diff
@@ -11,15 +11,10 @@ target_link_libraries(debug_op
...
-    os_regex
     os_net
-    string_op
     binaries_op
     regex_op
-    validate_op
-    expression
     randombytes
     bzip2_op
     cjson)
```
cborla, line 20: add a new line.

Issue #128, opened 2024-09-11
Create Installation Packages for Wazuh Agent


vikman90:
**Parent Issue:**
- https://github.com/wazuh/wazuh-agent/issues/126

## Description

This issue focuses on creating platform-specific packages (RPM, DEB, MSI, PKG) to facilitate the installation of the Wazuh agent across supported operating systems. The packages should integrate with the native package managers and follow best practices for each platform.

## Functional Requirements

- **Linux**: Create RPM and DEB packages to install the Wazuh agent on RHEL-based and Debian-based distributions.
- **Windows**: Create an MSI installer for the Wazuh agent.
- **macOS**: Create a PKG installer for macOS 14+.

## Subtasks

- [x] https://github.com/wazuh/wazuh-agent/issues/133
   - Develop an RPM spec file for RHEL, CentOS, Fedora, and similar distributions.
- [x] https://github.com/wazuh/wazuh-agent/issues/143
   - Create a DEB package for Debian-based distributions (Ubuntu, Debian, etc.).
- [x] https://github.com/wazuh/wazuh-agent/issues/134
   - Develop an MSI installer for Windows.
   - Ensure the installer correctly registers the agent as a Windows Service.
   - Handle dependencies and any required installation scripts (pre/post-install).
- [x] https://github.com/wazuh/wazuh-agent/issues/135
   - Develop a PKG installer for macOS.
   - Ensure the installer sets up the agent as a launch daemon.
   - Handle installation-related tasks like setting up permissions and configuring files.
- [x] **Documentation:**
   - Create installation instructions for each type of package (RPM, DEB, MSI, PKG).
   - Provide troubleshooting guidelines for package-related issues.

mjcr99:
## Update
**(16/10/2024)** Added option to run `make install` and let every built file in the desired folder based on the building environment. We are waiting for this development to complete the service installation:
* https://github.com/wazuh/wazuh-agent/pull/221

lchico:
# Update
- I have been working on the following issues:

```bash
[ 98%] Building CXX object CMakeFiles/wazuh-agent.dir/agent/src/main.cpp.o
[100%] Linking CXX executable wazuh-agent
[...]

PR #182, opened 2024-09-27
Remove entrypoints from Dockerfiles and create volumes for scripts

jotacarma90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/128|

## Description
In this PR we are going to include all the necessary changes so that the Dockerfile to generate agent packages does not contain directly any script inside the image, removing also the entrypoint.
This change is to eliminate the need to update the image every time a generation script is updated. This way, the Docker image generation workflows will only trigger when the Dockerfile itself is modified.

## Test
I was able to successfully create a package using branch `enhancement/128-remove-entrypoints`, and cherry-picking the commit of this PR. In this way, the Docker image used did not contain the scripts inside, but they were passed using a volume.

### DIFF
--- a/packages/generate_package.sh
+++ b/packages/generate_package.sh
@@ -27,6 +27,7 @@ CHECKSUM="no"
 FUTURE="no"
 LEGACY="no"
 IS_STAGE="no"
+ENTRYPOINT="/home/build.sh"
 
 
 trap ctrl_c INT
@@ -97,7 +98,7 @@ build_pkg() {
         ${CUSTOM_CODE_VOL} \
         -v ${DOCKERFILE_PATH}:/home:Z \
         ${CONTAINER_NAME}:${DOCKER_TAG} \
-        /home/build.sh \
+        ${ENTRYPOINT} \
         ${REVISION} ${JOBS} ${DEBUG} \
         ${CHECKSUM} ${FUTURE} ${LEGACY} ${SRC}|| return 1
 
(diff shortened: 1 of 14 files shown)


mjcr99 (changes requested):
We could also add the required changes in this [file](https://github.com/wazuh/wazuh-agent/blob/master/.github/workflows/packages_upload_images.yml).

Regarding this [file](https://github.com/wazuh/wazuh-agent/blob/master/.github/workflows/packages_retag_images.yml) we should update it accordingly since the [wazuh/wazuh](https://github.com/wazuh/wazuh/blob/master/.github/workflows/packages-retag-images.yml) version includes the following PR which solves a known problem:

* https://github.com/wazuh/wazuh/pull/24187

packages/generate_package.sh
```diff
@@ -95,7 +95,9 @@ build_pkg() {
         -e IS_STAGE="${IS_STAGE}" \
         -e WAZUH_BRANCH="${BRANCH}" \
         ${CUSTOM_CODE_VOL} \
+        -v ${DOCKERFILE_PATH}:/home:Z \
         ${CONTAINER_NAME}:${DOCKER_TAG} \
+        /home/build.sh \
```
MarcelKemp, line 100: Can we modify this line and turn it into a variable which we set earlier in the code?
Such as `${ENTRYPOINT}`
```suggestion
       ${ENTRYPOINT} \
```

MarcelKemp (changes requested):

packages/generate_package.sh
jotacarma90: Done

jotacarma90:
# Test
Package created using this branch merged with `enhancement/143-linux-deb-package-creation`:

```
Package wazuh-agent_5.0.0-0_amd64_466d96772.deb added to /home/vagrant/wazuh-agent.
+ return 0
+ return 0
+ clean 0
+ exit_code=0
+ find /home/vagrant/wazuh-agent/packages/debs/amd64/agent '(' -name '*.sh' -o -name '*.tar.gz' -o -name 'wazuh-*' ')' '!' -name docker_builder.sh -exec rm -rf '{}' +
+ exit 0
```

mjcr99 (approved):
LGTM

Issue #142, opened 2024-09-11
Implement Daemon Mode for Windows


vikman90:
**Parent Issue:**
- https://github.com/wazuh/wazuh-agent/issues/140

## Description

This task focuses on implementing the daemon (background service) mode of the Wazuh agent for Windows systems. The agent will need to function as a Windows service that can be managed using the standard Windows Service Manager interface.

## Functional Requirements

1. Implement the Wazuh agent as a Windows service using Windows Service API.
2. Create a logging mechanism that writes to the Windows Event Viewer.
3. Integrate the agent with the Windows Service Manager, enabling the following CLI commands:
   - `wazuh-agent start` (starts the agent service)
   - `wazuh-agent status` (checks if the agent is running as a service)
   - `wazuh-agent stop` (stops the agent service)
4. The agent should handle Windows service events and control signals gracefully.

## Non-Functional Requirements

- The service must be properly registered with the Windows Service Manager and adhere to Windows service best practices.
- Error handling and logs should be properly written to the Windows Event Viewer.

## Acceptance Criteria

1. The agent correctly responds to `start`, `stop`, and `status` commands.
2. Logs are appropriately stored in the Windows Event Viewer.

sdvendramini:
## 18/09/2024
Understanding issue requirements and doing research on windows services implementation. Looked at the old wazuh-agent windows service code. Synch with team mates to develop the parts this issue and the Unix daemon issue have in common.

PR #178, opened 2024-09-25
Implement wazuh agent daemon mode

aritosteles:
|Related issue|
|---|
|[141](https://github.com/wazuh/wazuh-agent/issues/141)|

## Description

This PR aims to make wazuh-agent run in the background and support CLI commands start, status and stop.

- Compilation without warnings in every supported platform
  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

PR #184, opened 2024-09-27
Implement Windows Service for the agent

sdvendramini:
|Related issue|
|---|
|[141](https://github.com/wazuh/wazuh-agent/issues/142)|

## Description

This PR aims to make wazuh-agent run in the background and support CLI commands start, status, restart and stop creating a windows service.

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,11 @@ add_subdirectory(agent)
 add_executable(wazuh-agent agent/src/main.cpp)
 target_link_libraries(wazuh-agent Agent Logger)
 
+if(WIN32)
+    include(cmake/SetWindowsManifest.cmake)
+    set_windows_manifest()
+endif()
+
 include(cmake/ConfigureTarget.cmake)
 configure_target(wazuh-agent)
 
--- a/src/agent/src/main.cpp
+++ b/src/agent/src/main.cpp
@@ -7,8 +7,6 @@ int main(int argc, char* argv[])
     Logger logger;
     CommandlineParser cmdParser(argc, argv);
 
-    LogInfo("Starting wazuh-agent service.");
-
     try
     {
         if (cmdParser.OptionExists("register"))
@@ -33,6 +31,20 @@ int main(int argc, char* argv[])
         else if (cmdParser.OptionExists("stop"))
         {
             StopAgent();
+        }
+        else if (cmdParser.OptionExists("install"))
+        {
+            if (!InstallService())
+                return 1;
+        }
+        else if (cmdParser.OptionExists("remove"))
+        {
+            if (!RemoveService())
+                return 1;
+        }
+        else if (cmdParser.OptionExists("service"))
+        {
+            SetDispatcherThread();
         }
         else
         {
--- a/src/agent/src/windows_service.cpp
+++ b/src/agent/src/windows_service.cpp
@@ -1,0 +1,176 @@
+#include <windows_service.hpp>
+
+#include <logger.hpp>
+
+#include <chrono>
+#include <thread>
+
+namespace
+{
+    SERVICE_STATUS g_ServiceStatus = {};
+    SERVICE_STATUS_HANDLE g_StatusHandle = nullptr;
+    HANDLE g_ServiceStopEvent = INVALID_HANDLE_VALUE;
+    SERVICE_DESCRIPTION g_serviceDescription;
+    const std::string AGENT_SERVICENAME = "Wazuh Agent";
+    const std::string AGENT_SERVICEDESCRIPTION = "Wazuh Windows Agent";
+
+    std::string GetExecutablePath()
+    {
+        char buffer[MAX_PATH];
+        GetModuleFileName(NULL, buffer, MAX_PATH);
+        return std::string(buffer);
+    }
+} // namespace
+
+namespace WindowsService
+{
+    bool InstallService()
+    {
+        const std::string exePath = GetExecutablePath() + " service";
+
+        SC_HANDLE schSCManager = OpenSCManager(nullptr, nullptr, SC_MANAGER_CREATE_SERVICE);
+        if (!schSCManager)
+        {
+            LogError("OpenSCManager fail: {}", GetLastError());
+            return false;
+        }
+
+        SC_HANDLE schService = CreateService(schSCManager,
+                                             AGENT_SERVICENAME.c_str(),
+                                             AGENT_SERVICENAME.c_str(),
+                                             SERVICE_ALL_ACCESS,
+                                             SERVICE_WIN32_OWN_PROCESS,
+                                             SERVICE_AUTO_START,
+                                             SERVICE_ERROR_NORMAL,
+                                             exePath.c_str(),
+                                             nullptr,
+                                             nullptr,
+                                             nullptr,
+                                             nullptr,
+                                             nullptr);
+
+        if (!schService)
+        {
+            LogError("CreateService fail: {}", GetLastError());
+            CloseServiceHandle(schSCManager);
+            return false;
+        }
+
+        g_serviceDescription.lpDescription = const_cast<char*>(AGENT_SERVICEDESCRIPTION.c_str());
+        ChangeServiceConfig2(schService, SERVICE_CONFIG_DESCRIPTION, &g_serviceDescription);
+
+        CloseServiceHandle(schService);
+        CloseServiceHandle(schSCManager);
+
+        LogInfo("Wazuh Agent Service successfully installed.");
+
+        return true;
+    }
+
+    bool RemoveService()
+    {
+        SC_HANDLE schSCManager = OpenSCManager(nullptr, nullptr, SC_MANAGER_CREATE_SERVICE);
+        if (!schSCManager)
+        {
+            LogError("OpenSCManager fail: {}", GetLastError());
+            return false;
+        }
+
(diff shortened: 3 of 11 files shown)


src/agent/src/windows_service.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+namespace WindowsService
+{
+
+    void InstallService()
+    {
+        const std::string exePath = GetExecutablePath() + " service";
+
+        SC_HANDLE schSCManager = OpenSCManager(nullptr, nullptr, SC_MANAGER_CREATE_SERVICE);
+        if (schSCManager == nullptr)
```
jr0me, line 80: There's a mix of styles when checking pointers in these files, sometimes comparing explicitly against `nullptr` and sometimes checking the implicit conversion to bool. Can we settle for the latter? (either one is ok...)

src/agent/include/process_options.hpp
```diff
@@ -10,7 +10,8 @@ void RestartAgent();
 void StartAgent();
 void StatusAgent();
 void StopAgent();
+void RestartAgent();
 void PrintHelp();
-void InstallService([[maybe_unused]] const std::string& exePath);
+void InstallService();
```
TomasTurina, line 15: Are these three functions necessary here?

src/agent/src/windows_service.cpp
```diff
@@ -19,6 +18,32 @@ namespace
         GetModuleFileName(NULL, buffer, MAX_PATH);
         return std::string(buffer);
     }
+
+    void HandleStopSignal(const char* logMessage, DWORD CtrlCode)
```
jr0me, line 24: Nitpick, should be `ctrlCode`.

src/agent/src/windows_service.cpp
```diff
@@ -44,11 +46,32 @@ namespace
             LogError("Failed to set service status to {}. Error: {}", g_ServiceStatus.dwCurrentState, GetLastError());
         }
     }
+
+    std::string ServiceStatusToString(DWORD dwCurrentState)
```
jr0me, line 50: Variable name could be `currentState`.

src/CMakeLists.txt
```diff
@@ -14,6 +14,17 @@ target_include_directories(Agent PUBLIC ${CMAKE_SOURCE_DIR}/include)
 add_executable(wazuh-agent agent/src/main.cpp)
 target_link_libraries(wazuh-agent Agent Logger)
 
+if(WIN32)
+    set(MANIFEST_FILE "${CMAKE_SOURCE_DIR}/app.manifest")
```
jr0me, line 18: This could be moved into a `src/cmake/SetWindowsManifest.cmake` file (name could be different), and reduce this to a function call. But we can also do this later if the cmake file keeps growing.

src/app.manifest
jr0me, line 1: I'm not sure that this file should be in the `src` directory. Was it like this in the old repository? Anyway, we can also move this later on...

src/agent/include/process_options.hpp
jr0me: I think that this is the most straightforward approach to keep the PR focus on the development of service/daemon. We could improve this on a separate issue, an alternative would be to make the command line parser platform dependent as well, and only recognize certain options on specific platforms.

src/agent/src/main.cpp
```diff
@@ -34,6 +32,20 @@ int main(int argc, char* argv[])
         {
             StopAgent();
         }
+        else if (cmdParser.OptionExists("install"))
+        {
+            InstallService();
+            return 0;
```
jr0me, line 38: Some comment for `RemoveService`, is there no way to tell if any of them failed? We always return 0, which is a success return value, if we can tell that installing or removing the service failed, we could return a non zero value.

src/agent/src/windows_service.cpp
```diff
@@ -0,0 +1,176 @@
...
+    {
+        char buffer[MAX_PATH];
+        GetModuleFileName(NULL, buffer, MAX_PATH);
+        return std::string(buffer);
+    }
+} // namespace
+
+namespace WindowsService
+{
+
```
jr0me: nitpicking here, remove extra line.

src/agent/src/windows_service.cpp
```diff
@@ -0,0 +1,176 @@
...
+            CloseServiceHandle(schSCManager);
+            return;
+        }
+
+        CloseServiceHandle(schService);
+        CloseServiceHandle(schSCManager);
+
+        LogInfo("Wazuh Agent Service successfully installed.");
+
+        return;
```
jr0me, line 112: This return is implied, same for lines 96, 97.

src/agent/src/windows_service.cpp
```diff
@@ -195,4 +217,103 @@ namespace WindowsService
...
+            return;
+        }
+
+        SC_HANDLE hService = OpenService(hSCManager, AGENT_SERVICENAME.c_str(), SERVICE_START);
+        if (!hService)
+        {
+            LogError("Error: Unable to open service. Error: {}", GetLastError());
+            CloseServiceHandle(hSCManager);
+            return;
+        }
```
jr0me, line 235: Up to here, this code is almost the same for `SeriviceStart|ServiceStop|ServiceStatus`. Since we have it 3 times, maybe we can extract a function from this.

jr0me (commented):
Looking good! Great to see the comments on each commit 💯 they help a lot to understand what's goin on, thanks!

I left a couple of comments, mostly opinionated stuff which I leave to your discretion!

src/CMakeLists.txt
sdvendramini: Done!

src/app.manifest
sdvendramini: In the old repository, these types of files are in the "win32" folder. Do you think it would be better if I create it and place the file inside?

src/agent/include/process_options.hpp
sdvendramini: Yes, we could improve this in the next one that completes the epic.

src/agent/src/main.cpp
sdvendramini: Fixed!

src/agent/src/windows_service.cpp
sdvendramini: done!

src/agent/src/windows_service.cpp
sdvendramini: No longer applies, as it now returns a bool!

src/agent/src/windows_service.cpp
sdvendramini: Fixed!

src/agent/src/windows_service.cpp
sdvendramini: Fixed!

src/agent/src/windows_service.cpp
sdvendramini: Done!

src/agent/src/windows_service.cpp
sdvendramini: Fixed! I decided to use `if (!varName)`.

src/app.manifest
jr0me: Yes, I think it could be good as a start to keep things separated. Though, will we need another file for the linux/macOS daemons too? If that's the case, maybe one folder for the manifest files, instead of a folder for windows only files could be better for the time being.

src/agent/src/windows_service.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include <agent.hpp>
+#include <logger.hpp>
+#include <signal_handler.hpp>
+
+#include <memory>
+#include <unordered_map>
+
+SERVICE_STATUS g_ServiceStatus = {};
+SERVICE_STATUS_HANDLE g_StatusHandle = nullptr;
+HANDLE g_ServiceStopEvent = INVALID_HANDLE_VALUE;
```
jr0me, line 12: I believe that these could go into the anonymous namespace, as a safety precaution.

src/agent/src/agent_registration.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+        nlohmann::json bodyJson = {{"id", m_agentInfo.GetUUID()}, {"key", m_agentInfo.GetKey()}};
+
+        if (!m_agentInfo.GetName().empty())
+        {
+            bodyJson["name"] = m_agentInfo.GetName();
+        }
+
+        const auto reqParams = http_client::HttpRequestParams(
+            http::verb::post, m_managerIp, m_managerPort, "/api/v1/agents", token.value(), "", bodyJson.dump());
```
Nicogp, line 51: Why do you change the endpoint to `api/v1/agents`? is this correct?

src/agent/src/windows_service.cpp
sdvendramini: Done!

src/agent/src/agent_registration.cpp
sdvendramini: This is wrong. The change will be reverted in [this PR](https://github.com/wazuh/wazuh-agent/pull/178).

src/app.manifest
sdvendramini: Finally I moved the file to  `agent/service/` folder.

jr0me (approved):

Nicogp (approved):

PR #187, opened 2024-09-30
Fix null pointer possible usage

ncvicchi:
## Description

If a malloc memory allocation failed in file_op.c, although the pointer returned was being checked for success, thhe process was allowed to continue, thus allowing to use a null pointer in several functions, allowing an undetermined behavior.


cborla (approved):
LGTM

Issue #158, opened 2024-09-13
Implement HTTPS support in the network wrapper


TomasTurina:
## Description

As part of the agent MVP development, a new communication protocol was developed that uses HTTP to communicate with the agent comms and server management APIs. For this purpose, the boost asio/beast library was selected.

As part of the new step, we need to implement HTTPS support using the boost networking library.

Tasks to be done:
- [x] Make a POC to test HTTPS boost library support.
- [x] Implement HTTPS support in a new HTTPSClient class.
- [x] Add an option to select between HTTPS and HTTP in the agent.
- [x] Make the communicator able to use both implementations.

Nicogp:
## Update 

(2024/09/17):

- I started researching about using https with Boost
- I started a first https PoC with Boost

PR #189, opened 2024-09-30
Allows communication using https

Nicogp:
|Related issue|
|---|
|#158|

## Description

This PR adds the changes to allow communication using https. It also adds a configuration option `“https_enabled”` to allow the use of http or https.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

### DIFF
--- a/src/agent/communicator/src/communicator.cpp
+++ b/src/agent/communicator/src/communicator.cpp
@@ -41,8 +41,9 @@ namespace communicator
         {
             m_managerIp = getStringConfigValue("agent", "manager_ip");
             m_port = getStringConfigValue("agent", "agent_comms_api_port");
-            const std::string httpsEnabled =  getStringConfigValue("agent", "https_enabled");
-            if (httpsEnabled == "no") {
+            const std::string httpsEnabled = getStringConfigValue("agent", "https_enabled");
+            if (httpsEnabled == "no")
+            {
                 m_useHttps = false;
                 LogInfo("Using insecure connection.");
             }
@@ -100,8 +101,8 @@ namespace communicator
             return m_keepRunning.load();
         };
 
-        const auto reqParams =
-            http_client::HttpRequestParams(boost::beast::http::verb::get, m_managerIp, m_port, "/api/v1/commands", m_useHttps);
+        const auto reqParams = http_client::HttpRequestParams(
+            boost::beast::http::verb::get, m_managerIp, m_port, "/api/v1/commands", m_useHttps);
         co_await m_httpClient->Co_PerformHttpRequest(
             m_token, reqParams, {}, onAuthenticationFailed, onSuccess, loopCondition);
     }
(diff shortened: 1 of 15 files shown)


src/agent/communicator/src/http_socket_factory.hpp
```diff
@@ -15,7 +15,8 @@ namespace http_client
     public:
         std::unique_ptr<IHttpSocket> Create(const boost::asio::any_io_executor& executor) override
         {
-            return std::make_unique<HttpSocket>(executor);
+            //return std::make_unique<HttpSocket>(executor);
+            return std::make_unique<HttpsSocket>(executor);
```
jr0me: Looks like these changes were made to make some local tests, can they be reverted?

src/agent/communicator/src/communicator.cpp
```diff
@@ -41,12 +41,17 @@ namespace communicator
         {
             m_managerIp = getStringConfigValue("agent", "manager_ip");
             m_port = getStringConfigValue("agent", "agent_comms_api_port");
+            std::string strHttps =  getStringConfigValue("agent", "https_enabled");
```
jr0me, line 44: The string can be const. Also, since the type is already visible in this scope, there's no need for using Hungarian notation, it could be simply named "httpsEnabled" or "isHttpsEnabled".

src/agent/src/register.cpp
```diff
@@ -16,9 +16,10 @@ namespace registration
         const configuration::ConfigurationParser configurationParser;
         const auto managerIp = configurationParser.GetConfig<std::string>("agent", "manager_ip");
         const auto managerPort = configurationParser.GetConfig<std::string>("agent", "server_mgmt_api_port");
+        const bool useHttps = (configurationParser.GetConfig<std::string>("agent", "https_enabled") == "no") ? false : true;
```
jr0me: This could be:

```
const bool useHttps = !(configurationParser.GetConfig<std::string>("agent", "https_enabled") == "no");
```

src/agent/communicator/src/http_socket.hpp
```diff
@@ -60,67 +59,4 @@ namespace http_client
     private:
         boost::asio::ip::tcp::socket m_socket;
     };
-
-    class HttpsSocket : public IHttpSocket
```
jr0me: These changes could be squashed to the first commit, so the new class is introduced from the beginning in a separate file.

Nicogp: Done!

src/agent/src/register.cpp
Nicogp: Done!

src/agent/communicator/src/communicator.cpp
Nicogp: Done!

src/agent/communicator/src/http_socket_factory.hpp
Nicogp: Done!

TomasTurina (approved):

jr0me (approved):

Issue #146, opened 2024-09-11
Integrate the Command Handler with the Module Manager


TomasTurina:
**Parent issue:**
- https://github.com/wazuh/wazuh/issues/22887

## Description

After the development of the new [Command Handler](https://github.com/wazuh/wazuh-agent/issues/83), it is necessary to integrate it with the Module Manager so that modules can execute commands and report their feedback.

## Considerations

- All commands should be executed in the order they are created.
- The Command Handler should use the Module Manager to find a given module and send the command.
- A coroutine should be created for each command to wait for the command to be executed and for the module to respond.
- The command response should be stored in the Queue so that the result is communicated to the manager.
- The command response should be persisted in the command store.
- If no response is received after a time, the coroutine should consider marking the command status as timeout.
  - Result should also be stored in the Queue.

After the implementation is complete, we also need to add some UT to test the entire execution flow of a command.

jr0me:
### Update

Started looking into the issue and its requirements. Also looking at the current state of the involved components' code and the branch enhancement/17-migrate-syscollector-to-inventory-at-new-agent.

jr0me:
### Update

Created branch and linked it to this issue. Started looking into the `CommandHandler`, `CommandStore` and `ModuleWrapper` to come up with a solution.

jr0me:
### Update

Implemented first attempt to process a command, fetching its corresponding module and passing to it the command b4dc05b0c357c08ca16f8449efa0a554f401c325

Now it's necessary to come up with a non blocking mechanism to execute commands.

jr0me:
### Update

Implemented first approach with coroutines. The functions (in modules, module wrapper, etc) that execute the commands needed to return `awaitables` among other changes. Testing with the new server comms mvp.

jr0me:
### Update

Defined an initial format for the stateful messages created by command executions.

```json
{
  "data":
  {
    "id": str,
    "error": int,
    "message": str
  },
  "module": str
}
```

Further improvements to the coroutines approach used so far. There's a dependency between the ModularManager, ModuleWrapper, Inventory, Command, CommandStore, CommandHandler that needs further iterations to improve.

jr0me:
### Update

Created draft PR. Timeout implementation is currently missing, as well as the insurance of execution order, since at the moment only one sequential execution at the time is allowed.

jr0me:
### Update

[...]

PR #202, opened 2024-10-04
Support simultaneous command dispatch with timeout

jr0me:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/146|

## Description

In this PR we add support for simultaneous command dispatch and timeouts.

The timeout on each command dispatch is accomplish by launching a timer coroutine alongside the command coroutine. Multiple simultaneous command dispatch is done by spawning coroutines with `co_spawn`.

  - [x] Linux
  - [ ] Windows
  - [ ] MAC OS X

### DIFF
--- a/src/agent/src/command_handler_utils.cpp
+++ b/src/agent/src/command_handler_utils.cpp
@@ -29,9 +29,17 @@ DispatchCommand(module_command::CommandEntry commandEntry,
     // NOLINTNEXTLINE(cppcoreguidelines-avoid-capturing-lambda-coroutines)
     auto executeCommandTask = [module, commandEntry, result, commandCompleted, timer]() -> boost::asio::awaitable<void>
     {
-        *result = co_await module->ExecuteCommand(commandEntry.Command);
-        *commandCompleted = true;
-        timer->cancel();
+        try
+        {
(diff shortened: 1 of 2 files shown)


src/agent/src/command_handler_utils.cpp
```diff
@@ -15,15 +78,23 @@ DispatchCommand(module_command::CommandEntry commandEntry,
 
     LogInfo("Dispatching command {}({})", commandEntry.Command, commandEntry.Module);
 
-    const auto result = co_await module->ExecuteCommand(commandEntry.Command);
+    const auto timeout = std::chrono::minutes(60);
```
Nicogp, line 81: Maybe we should analyze if we need to make the timeout more flexible, for example by configuration or even that it can come as data from the api. But I think that may be for another issue.

jr0me: Yes, totally, we have many timers with hardcoded values atm. We need to address them with some customizable configuration.

sdvendramini (approved):

Nicogp (approved):

cborla (approved):
LGTM

Issue #207, opened 2024-10-08
Implement Selective Compilation of Optional Modules


cborla:
## **Description**

Currently the components are crucial for the application's functionality, while the modules are libraries that extend functionality by utilizing the essential components.

The goal of this issue is to implement **selective compilation** that allows including or excluding optional modules during the build process, using **CMake** to manage this configuration efficiently.

---

## **Motivation**

- **Flexibility:** Allow users to customize the application based on their needs by including only the required modules.
- **Optimization:** Reduce the final size of the application by excluding unnecessary modules.
- **Maintenance:** Facilitate the management and updating of optional modules without affecting the core application.

## **Proposed Solution**
The default behaviour is to always compile including all modules, and in case you want to exclude the compilation of one of them you can do it by command line or by disabling the internal define in cmake.
```bash
cmake -DUSE_FIM=OFF -DUSE_INVENTORY=OFF ..
```

## **Tasks**
1. Update CMakeLists.txt:
     - [x] Implement conditional options for modules.
     - [x] Configure dependencies between components and modules appropriately.
2. Modify Source Code:
    - [x] Add preprocessor macros where necessary to manage module-specific code inclusion, the idea is to only include the safeguards at one point, preferably in the `modules manager`.
3. Testing:
    - [x] Verify builds with different combinations of modules enabled and disabled.
    - [x] Ensure the application functions correctly in all configurations.
4. Documentation:
    - [ ] Update project documentation to reflect the new build options.
    - [ ] Include instructions on how to disable modules during the build process, include options in the `BUILD.md` file.

nbertoldo:
### Update 2024/10/08

- Implement conditional options for modules:

```Cmake
option(ENABLE_MODULE_INVENTORY "Enable module Inventory" OFF)

if(ENABLE_MODULE_INVENTORY)
    message(STATUS "Module Inventory is enabled")
    add_definitions(-DENABLE_MODULE_INVENTORY)
    add_subdirectory(inventory)
else()
    message(STATUS "Module Inventory is disabled")
endif()
```

nbertoldo:
### Update 2024/10/09

- Refactoring _ModuleManager_ to allow selective module building.

nbertoldo:
### Update 2024/10/14

- Integrate _taskManager_ to _ModuleManager_.
- Suggested changes.

nbertoldo:
### Update 2024/10/15

- Refactor _ModuleManager_.
- Rebase to _master_.
[...]

PR #213, opened 2024-10-10
Optional module compilation

nbertoldo:
|Related issue|
|---|
| #207 |

## Description

This PR implements selective compilation that allows including or excluding optional modules during the build process, using CMake to manage this configuration efficiently.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,9 +13,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     add_compile_options(/w)
 endif()
 
-# Enable/disable optional modules compilation
-option(ENABLE_MODULE_INVENTORY "Enable module Inventory" ON)
-
 add_subdirectory(common)
 add_subdirectory(modules)
 add_subdirectory(agent)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -13,6 +13,9 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     add_compile_options(/w)
 endif()
 
+# Enable/disable optional modules compilation
+option(ENABLE_MODULE_INVENTORY "Enable module Inventory" ON)
+
 add_subdirectory(common)
 add_subdirectory(modules)
 add_subdirectory(agent)
--- a/src/agent/src/agent.cpp
+++ b/src/agent/src/agent.cpp
@@ -28,7 +28,9 @@ Agent::Agent(const std::string& configPath, std::unique_ptr<ISignalHandler> sign
                      m_agentInfo.GetKey(),
                      [this](std::string table, std::string key) -> std::string
                      { return m_configurationParser.GetConfig<std::string>(std::move(table), std::move(key)); })
-    , m_moduleManager(m_messageQueue, m_configurationParser)
+    , m_moduleManager([this](Message message) -> int { return m_messageQueue->push(std::move(message)); },
+                      m_configurationParser,
+                      [this](std::function<void()> task) { m_taskManager.EnqueueTask(std::move(task)); })
 {
     m_centralizedConfiguration.SetGroupIdFunction([this](const std::vector<std::string>& groups)
                                                   { return m_agentInfo.SetGroups(groups); });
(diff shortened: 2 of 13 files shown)


src/CMakeLists.txt
```diff
@@ -13,6 +13,9 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
     add_compile_options(/w)
 endif()
 
+# Enable/disable optional modules compilation
```
jr0me, line 16: Since this option is related to the ModuleManager, could we move it to its corresponding cmake?

src/modules/include/moduleManager.hpp
```diff
@@ -53,6 +53,7 @@ class ModuleManager {
     }
 
     std::shared_ptr<ModuleWrapper> GetModule(const std::string & name);
+    void Init();
```
jr0me, line 56: Do we really need to have separated Init and Setup functions? The current code doesn't seem to ask for it... maybe I'm missing the big picture but looks like the additional complexity is not needed at the moment. Could something be moved to the constructor?

src/CMakeLists.txt
nbertoldo: Done

src/modules/include/moduleManager.hpp
nbertoldo: Done

cborla:
Moved to on hold until the release testing stage is completed.

cborla (approved):
LGTM

src/modules/include/moduleManager.hpp
```diff
@@ -11,21 +11,23 @@
...
-                             const std::shared_ptr<IMultiTypeQueue> queue) {
+                             const std::function<int(Message)>& pushMessage) {
     { t.Start() } -> std::same_as<void>;
     { t.Setup(configurationParser) } -> std::same_as<void>;
     { t.Stop() } -> std::same_as<void>;
     { t.ExecuteCommand(query) } -> std::same_as<std::string>;
     { t.Name() } -> std::same_as<std::string>;
-    { t.SetMessageQueue(queue) } -> std::same_as<void>;
+    { t.SetPushMessageFunction(pushMessage) } -> std::same_as<void>;
 };
```
jr0me, line 21: We are not using this, right? I would remove it if so, it's a problem to maintain code that it is not being used.

src/modules/CMakeLists.txt
```diff
@@ -6,24 +6,36 @@ include(../cmake/CommonSettings.cmake)
...
+    target_link_libraries(ModuleManager PRIVATE Inventory)
+    add_subdirectory(inventory)
+endif()
 
 if (ENABLE_LOGCOLLECTOR)
+    add_subdirectory(logcollector)
     target_link_libraries(ModuleManager PUBLIC Logcollector)
 endif()
 
+target_link_libraries(ModuleManager PUBLIC ConfigurationParser MultiTypeQueue Logger ModuleCommand)
```
jr0me, line 30: It looks like `Logger` and `ModuleCommand` could be linked privately.

src/modules/inventory/include/inventory.hpp
```diff
@@ -31,7 +31,7 @@ class Inventory {
         void Stop();
         Co_CommandExecutionResult ExecuteCommand(const std::string query);
         const std::string& Name() const { return m_moduleName; };
-        void SetMessageQueue(const std::shared_ptr<IMultiTypeQueue> queue);
+        void SetPushMessageFunction(const std::function<int(Message)>& pushMessage);
```
jr0me, line 34: The return type could also be a `bool` right? Maybe the idea is to support error codes in the future? I would still then define an enum class for that, though in the meantime I'd prefer to narrow it down.

src/agent/src/agent.cpp
```diff
@@ -71,14 +65,7 @@ void Agent::Run()
...
 
-    m_moduleManager.AddModule(Inventory::Instance());
-
-#ifdef ENABLE_LOGCOLLECTOR
-    m_moduleManager.AddModule(Logcollector::Instance());
-#endif
-
-    m_moduleManager.AddModule(m_centralizedConfiguration);
-    m_moduleManager.Setup();
+    m_moduleManager.AddModules();
```
jr0me, line 68: I realize now that it would be best to add the modules and enqueue the module manager task before queuing the `CommandsProcessingTask` because the latter needs the former. Picture this, if the `AddModules` and `Start` takes long enough, a command could try to be pushed into an unitialized, or unstarted, module.

jr0me (approved):
LGTM. With just a few notes...

src/modules/inventory/include/inventory.hpp
nbertoldo: The return type is `<int>` since the _push_ function of the queue has that return type. In the future we should define error codes.

PR #217, opened 2024-10-14
Add libbz2-dev as a dependency in BUILD.md

jr0me:
## Description

A library is missing as a dependency in `BUILD.md`. Since it's not included in every distribution, we need to add it.


aritosteles (approved):

cborla (approved):
LGTM

PR #218, opened 2024-10-14
Update coverage check workflow

jr0me:
### Description

Github runners were updated to Ubuntu 24, these new runners already include LLVM (and clang) 16 by default. There's no need to manually install a specific version of LLVM. Additionally, since there's now an option to disable clang-tidy checks, we disable them for the coverage checks to speed it up -it's safe to do so, the static analysis already is covered by another workflows-. This PR also enables vcpkg's binary caching, bringing the running time to less than half of what it was previously.


cborla (approved):
LGTM

Issue #130, opened 2024-09-11
Create Systemd Service for Linux


vikman90:
**Parent Issue:**
- https://github.com/wazuh/wazuh-agent/issues/127

## Description

This subtask focuses on developing a systemd service for the Wazuh agent on Linux platforms. The service must ensure the agent starts automatically at boot, runs in the background, and can be managed via common systemd commands.

## Functional Requirements

- The agent service must be disabled by default.
- Ensure the service can be installed, enabled, started, stopped, and restarted via `systemctl`.
- Properly handle logging of service lifecycle events (start, stop, restart).

## Acceptance Criteria
- The Wazuh agent service is installed and enabled to start at boot.
- Administrators can control the service using standard systemd commands.
- Logs for service events are properly generated and accessible.

aritosteles:
# Update

- **2024/10/08**:
Research on sevice installation with systemd.
Added test for UnixDaemon class.

aritosteles:
##   Linux Options:

--run: Start wazuh-agent in the foreground.
--start: Start wazuh-agent daemon.
--status: Get wazuh-agent daemon status.
--stop: Stop wazuh-agent daemon.
--restart: Restart wazuh-agent daemon.
--register-agent: Register wazuh-agent.
--help: Print help message.

PR #221, opened 2024-10-15
Enhancement/130 create systemd service linux

aritosteles:
|Related issue|
|---|
|[130](https://github.com/wazuh/wazuh-agent/issues/130)|

## Description
This subtask focuses on developing a systemd service for the Wazuh agent on Linux platforms

  - [x] Linux


BUILD.md
```diff
@@ -60,10 +60,23 @@ The following dependencies are required for this project:
     cmake --build build
     ```
 5. **Run the Agent**
-    To run the agent, copy the file `src/agent/service/wazuh-agent.service` to `/etc/systemd/system/`.
-    Set the environment variable WAZUH_HOME to the wazuh-agent executable directory.
-    You can start and stop the agent, and get status with:
+    To run the agent, copy the file `src/agent/service/wazuh-agent.service` to `/usr/lib/systemd/system/`.
```
sdvendramini, line 63: Can we still run the agent from the CLI without creating the service? If so, perhaps we can add two items here:
- Run the Agent from the CLI
- Run the Agent as a service

Or something similar.

src/agent/tests/CMakeLists.txt
```diff
@@ -37,3 +37,9 @@ configure_target(cmd_ln_parser_test)
 target_include_directories(cmd_ln_parser_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include)
 target_link_libraries(cmd_ln_parser_test PRIVATE GTest::gtest)
 add_test(NAME CommandlineParserTest COMMAND cmd_ln_parser_test)
+
+add_executable(unix_daemon_test unix_daemon_test.cpp)
+configure_target(unix_daemon_test)
+target_include_directories(unix_daemon_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/../include ${CMAKE_CURRENT_SOURCE_DIR}/../src)
+target_link_libraries(unix_daemon_test PRIVATE Agent GTest::gtest)
+add_test(NAME UnixDaemonTest COMMAND unix_daemon_test)
```
sdvendramini, line 45: It looks like there's a new line missing at the end of the file. Could you add one, please?

BUILD.md
jr0me: I think it would be a good idea to allow, at least in debug mode, for the agent to run directly from the cli without needing to install a service.

src/common/logger/src/logger_linux.cpp
```diff
@@ -1,11 +1,11 @@
 #include <logger.hpp>
 
-#include <spdlog/sinks/systemd_sink.h>
+#include "spdlog/sinks/stdout_color_sinks.h"
```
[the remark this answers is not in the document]
aritosteles: Fixed.
Yes, we don´t need systemd right now, when the agent runs in the background systemd captures all output.

src/agent/tests/CMakeLists.txt
aritosteles: Fixed

BUILD.md
aritosteles: Yes, this came up in conversation with @vikman90. I just added this to BUILD.md.
`--run` runs the agent in the foreground, with all output sent to stdout
`--start `starts the service

sdvendramini (approved):

Issue #219, opened 2024-10-15
Agent centralized: necessary changes to existing modules


Nicogp:
**Parent issue:**

- https://github.com/wazuh/wazuh-agent/issues/32

### Description

As part of https://github.com/wazuh/wazuh-agent/issues/32 development it is necessary to add functionality to some of the existing modules:

- AgentInfo: it will persist the information of the group to which the agent belongs.
  - The group information will be stored in a new table, agent_group.
  - A SetGroups method will be added which will store the groups inside the table. When SetGroups is invoked, if they already have previous groups they must be removed and the new groups must be set.
  - A GetGroups method will be added that will read the database and return a list of the groups to which the agent belongs.

- Communicator: it will be in charge of making requests to the server to download the centralized configuration files.
  - Existing methods must be adapted or a new method must be created to perform the file download request.

Nicogp:
(2024/10/14):

- The corresponding changes were made to the AgentInfo module.
- Changes are being made to the communicator module.
  - I made a PoC to download files using http: successful result.
  - I made a PoC to download files using https: at the moment I am able to download the files but it is still giving an error, I am analyzing if it is because of how it is downloaded or a failure in the server I set up to perform the test.

PR #223, opened 2024-10-15
Handle agent centralized configuration commands

jr0me:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/222|

## Description

This PR introduces a CentralizedConfiguration class that will handle the incoming `set-group` and `update-group` commands.

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

PR #224, opened 2024-10-16
Add changes to persist agent_group information and perform file download requests

Nicogp:
|Related issue|
|---|
|#219|

## Description

This PR adds the changes to persist the groups to which the agent belongs and also adds the changes to HttpClient to be able to perform file download requests.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X


src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -75,6 +86,22 @@ void AgentInfoPersistance::CreateAgentInfoTable()
     }
 }
 
+void AgentInfoPersistance::CreateAgentGroupTable()
+{
+    try
+    {
+        m_db->exec("CREATE TABLE IF NOT EXISTS " + AGENT_GROUP_TABLE_NAME +
```
sdvendramini, line 93: I think you might be able to use `AGENT_INFO_TABLE_NAME` in the `CreateAgentInfoTable()` function, similar to how `AGENT_GROUP_TABLE_NAME` is used here.

src/agent/agent_info/include/agent_info.hpp
```diff
@@ -11,13 +12,16 @@ class AgentInfo
     std::string GetName() const;
     std::string GetKey() const;
     std::string GetUUID() const;
+    std::vector<std::string> GetGroups() const;
 
     void SetName(const std::string& name);
     void SetKey(const std::string& key);
     void SetUUID(const std::string& uuid);
+    void SetGroups(const std::vector<std::string>& group_list);
```
jr0me, line 20: Prefer camelCase for function parameters.

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -75,6 +86,22 @@ void AgentInfoPersistance::CreateAgentInfoTable()
     }
 }
 
+void AgentInfoPersistance::CreateAgentGroupTable()
```
jr0me, line 89: Do we need a table for the agent group? Can't it be just another column in the `agent_info` table?

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -127,6 +154,52 @@ std::string AgentInfoPersistance::GetAgentInfoValue(const std::string& column) c
...
+    }
+    catch (const std::exception& e)
+    {
+        LogError("Error inserting group: {}.", e.what());
+    }
+}
+
+std::vector<std::string> AgentInfoPersistance::GetAgentGroupList() const
+{
+    std::vector<std::string> group_list;
```
jr0me, line 184: Prefer camelCase for local variables too.

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -157,11 +235,17 @@ void AgentInfoPersistance::SetUUID(const std::string& uuid)
     SetAgentInfoValue("uuid", uuid);
 }
 
+void AgentInfoPersistance::SetGroups(const std::vector<std::string>& group_list)
+{
+    SetAgentGroups(group_list);
```
jr0me, line 240: Why don't just set the groups in this function instead of calling another one?

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -142,6 +215,11 @@ std::string AgentInfoPersistance::GetUUID() const
     return GetAgentInfoValue("uuid");
 }
 
+std::vector<std::string> AgentInfoPersistance::GetGroups() const
+{
+    return GetAgentGroupList();
```
jr0me, line 220: Why don't just get the groups in this function instead of calling another one?

src/agent/communicator/include/ihttp_socket.hpp
```diff
@@ -23,6 +23,9 @@ namespace http_client
 
         virtual void Read(boost::beast::http::response<boost::beast::http::dynamic_body>& res) = 0;
 
+        virtual void ReadToFile(boost::beast::http::response_parser<boost::beast::http::dynamic_body>& res,
```
jr0me, line 26: Wouldn't it be simpler to return a string? Instead of passing a reference. I'm also confused about the type and name for the parameter here, if it is an output file, shouldn't it be mutable? Why is it const?

src/agent/communicator/src/http_socket.hpp
```diff
@@ -43,6 +45,56 @@ namespace http_client
             boost::beast::http::read(m_socket, buffer, res);
         }
 
+        void ReadToFile(boost::beast::http::response_parser<boost::beast::http::dynamic_body>& res,
```
jr0me, line 48: This implementation and the one in `HttpsSocket` are almost identical, how can we avoid this code duplication?

jr0me:
We should add to this PR passing the new functions to the `CentralizedConfiguration` via its methods:

```c++
void SetGroupIdFunction(SetGroupIdFunctionType setGroupIdFunction);
void GetGroupIdFunction(GetGroupIdFunctionType getGroupIdFunction);
void SetDownloadGroupFilesFunction(DownloadGroupFilesFunctionType downloadGroupFilesFunction);
```

src/agent/agent_info/include/agent_info.hpp
Nicogp: Done!

src/agent/agent_info/src/agent_info_persistance.cpp
Nicogp: I think that in the future more information related to the groups could be added, for example the hash of the configuration file of each group to check that it has not been modified. That's why I thought about adding a new table.

src/agent/agent_info/src/agent_info_persistance.cpp
Nicogp: Done!

src/agent/agent_info/src/agent_info_persistance.cpp
Nicogp: Done!

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -157,11 +205,37 @@ void AgentInfoPersistance::SetUUID(const std::string& uuid)
     SetAgentInfoValue("uuid", uuid);
 }
 
+void AgentInfoPersistance::SetGroups(const std::vector<std::string>& groupList)
+{
+    try
+    {
+        SQLite::Transaction transaction(*m_db);
+
+        m_db->exec("DELETE FROM " + AGENT_GROUP_TABLE_NAME + ";");
```
jr0me, line 214: We should test this part too, for instance, set the groups to 1,2, verify, and then set to 2,3 and verify that 1 is no longer there.

src/agent/communicator/src/http_socket.hpp
Nicogp: Done!

src/agent/communicator/include/ihttp_socket.hpp
Nicogp: I changed the name of the parameter, this indicates the path and name of the file where it will be saved.

src/agent/agent_info/src/agent_info_persistance.cpp
Nicogp: Done!

src/agent/agent_info/src/agent_info_persistance.cpp
Nicogp: Done!

Nicogp:
In reference to this comment:

> We should add to this PR passing the new functions to the `CentralizedConfiguration` via its methods:
> 
> ```c++
> void SetGroupIdFunction(SetGroupIdFunctionType setGroupIdFunction);
> void GetGroupIdFunction(GetGroupIdFunctionType getGroupIdFunction);
> void SetDownloadGroupFilesFunction(DownloadGroupFilesFunctionType downloadGroupFilesFunction);
> ```
> 
> This can be done somewhere in `agent.cpp` either in its constructor or before adding it to the `ModuleManager`.


These changes will be added in PR [#223](https://github.com/wazuh/wazuh-agent/pull/223).

jr0me (approved):

sdvendramini (approved):

cborla (approved):
LGTM

PR #228, opened 2024-10-17
MVP for Logcollector Module

### LINKED ISSUE #201: Add Logcollector Module to Wazuh Agent
## Description

Logcollector is the agent module that adds the ability to collect system logs, both by reading plain text files and by collecting messages from the operating system API.

This task consists of redesigning the Logcollector module of the 4.x agent and implementing it in the new agent.

## Functional Requirements

- **Implement collectors** for:
    - Plaintext logs
    - macOS (using ULS) - macOS only
    - journald - Linux only
    - Windows (using eventchannel) - Windows only
- **Detect changes** including:
    - Wildcard matches for log file names
    - Rotated log files
    - Truncated log files
- **Store persistent state** for:
    - Reading position of each log file
    - Last modification date of each file
    - On start, reload state and act accordingly
- **Transmit logs** as stateless messages

## Non-functional Requirements

- A default configuration must be provided in the code.
- A proposal for configuration settings should be defined.

## Implementation Constraints

- Logcollector should be implemented as a module rather than a standalone executable.

## Deliverables

1. Wazuh Agent executable with the Logcollector module included
2. Design documentation for the module
3. Configuration documentation

vikman90:
**Related Issue:**
- Closes #201

## Summary

This PR introduces the Minimum Viable Product (MVP) for the Logcollector module in the Wazuh Agent. The module implements the ability to collect logs from local files, with support for concurrency, wildcard expansion, and file rotation detection.

## Features

- **Localfile block**: Reads local files line by line.
- **File rotation detection**: Detects rotated log files and continues reading.
- **Wildcard expansion**: Supports expanding wildcards in file paths.
- **Concurrency**: Log files are read concurrently, improving performance.

## Configuration Options

The module can be configured in the `wazuh.conf` file with the following options:

- `enabled`: Enables the Logcollector module.
- `localfiles`: List of files to monitor (supports wildcards).
- `reload_interval`: Interval to reload files and expand wildcards.
- `file_wait`: Interval between file reading cycles.

By default, Logcollector reads _/var/log/auth.log_.

## Build Options

To include the Logcollector module in the Wazuh Agent, compile with the following CMake option:

```bash
cmake -DENABLE_LOGCOLLECTOR=ON
```

The `ENABLE_LOGCOLLECTOR` option enables the Logcollector module and integrates it into the agent. That's enabled by default.

## Known Issues

- **Windows**: Logcollector-loaded files cannot be externally deleted while being monitored.
- **Linux**: If the agent forks, open file descriptors are copied, potentially leading to duplicated reads.

## Tested Operating Systems

- **Ubuntu 24.04**
- **macOS 15 Sequoia**
- **Windows 11**

---

Please review the implementation and provide any feedback for further improvement.


src/modules/logcollector/src/file_reader_unix.cpp
```diff
@@ -0,0 +1,34 @@
+#include "file_reader.hpp"
+#include <logcollector.hpp>
+#include <logger.hpp>
+#include <glob.h>
+
```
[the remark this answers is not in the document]
nbertoldo: Done

src/modules/logcollector/src/logcollector.cpp
```diff
@@ -0,0 +1,103 @@
+#include <logcollector.hpp>
+#include <logger.hpp>
+
+#include <boost/asio/co_spawn.hpp>
+#include <boost/asio/detached.hpp>
+
+#include "file_reader.hpp"
+
+using namespace logcollector;
+using namespace std;
```
LucioDonda, line 10: avoid `using namespace std;`

src/modules/logcollector/tests/unit/file_reader_test.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include <spdlog/spdlog.h>
+
+#include <logcollector.hpp>
+#include <file_reader.hpp>
+#include "queue_mock.hpp"
+#include "tempfile.hpp"
+#include "logcollector_mock.hpp"
+
+using namespace logcollector;
+using namespace std;
```
LucioDonda, line 13: delete line and replace all the occurrences with `std::<>`

src/modules/logcollector/tests/unit/logcollector_test.cpp
```diff
@@ -0,0 +1,43 @@
+#include <gtest/gtest.h>
+#include "logcollector_mock.hpp"
+#include <configuration_parser.hpp>
+#include <file_reader.hpp>
+#include "tempfile.hpp"
+
+using namespace configuration;
+using namespace logcollector;
+using namespace std;
```
LucioDonda, line 9: same here.

src/modules/logcollector/src/file_reader_unix.cpp
```diff
@@ -0,0 +1,35 @@
+#include "file_reader.hpp"
+#include <logcollector.hpp>
+#include <logger.hpp>
+#include <glob.h>
+#include <span>
+
+using namespace logcollector;
+using namespace std;
```
LucioDonda, line 8: same.

src/modules/logcollector/src/file_reader_win.cpp
```diff
@@ -0,0 +1,36 @@
+#include "file_reader.hpp"
+#include <logcollector.hpp>
+#include <logger.hpp>
+
+#include <windows.h>
+#include <string>
+#include <list>
+
+using namespace logcollector;
+using namespace std;
```
LucioDonda, line 10: same.

src/modules/logcollector/tests/unit/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+find_package(GTest CONFIG REQUIRED)
+
+FILE(GLOB LOGCOLLECTOR_TEST_SOURCES *_test.cpp)
+
+add_executable(logcollector_unit_tests ${LOGCOLLECTOR_TEST_SOURCES})
```
LucioDonda, line 5: There's no run all test main function

LucioDonda (changes requested):

nbertoldo (approved):

LucioDonda (approved):
LGTM!

cborla (approved):
LGTM

jr0me (commented):
The mvp looks good to me. There are a couple of notes that I would suggest to keep for future improvements:

* we should find a way to use clang-format in new files, even if we are in the folder structure that includes the old agent
* fix const correctness, there are many local variables that could be const, this improves readability and may improve compiler optimizations
* uses of `std::list`, I don't see the benefit of using it over `vector`, at least in these cases
* Logcollector has some virtual methods, but no derived classes. Seems to be a workaround for mocking it in tests, this could be improved.
* There are some uses of raw pointers that should be revised to use smart pointers if possible
* The use of references in coroutines is discouraged, it's preferable to rely in value semantics if possible
* Some directories were included publicly and maybe are not needed, same for libraries linked.
* There are missing headers, see `iwyu`.
* Some files introduce more than one class, each class should have its own file
* The `SetMessageQueue` introduced by the ModuleWrapper probably shouldn't use a *const* shared pointer. If it were mutable, it could be moved in each implementation of this function, I understand that that is the intended use of it
* Some `while(true)` have no exit points.

Issue #231, opened 2024-10-18
Add the '--config-path' option to be used when registering the agent.


Nicogp:
## Description

As part of the test of #127 ([PR](https://github.com/wazuh/wazuh-agent/pull/229)) we noticed the need to add the `--config-path` option to be used when registering the agent, otherwise it looks for a file named `wazuh.conf` in the execution directory or otherwise loads the default settings which in the case of `manager_ip `is `"localhost"`.

Example of how the registration would be performed with config-path option:

`./wazuh-agent --register-agent --user user --password pass --key key --config-path /path/to/file`

Nicogp:
(2024/10/18)

- I have performed manual tests
- I created this problem
- I made the necessary changes

PR #232, opened 2024-10-18
Added the '--config-path' option for '--register-agent'

Nicogp:
|Related issue|
|---|
|#231|

## Description

This PR adds the --config-path option so that it can also be used when registering the agent.

It also adds the following fixes:

- Reorder the **Run the Agent** section in BUILD.md
- Add the --config-path option to PrinHelp function
- And adds a fix to prevent the agent from trying to read the “” file when the --config-path option is not used.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X

### DIFF
--- a/src/agent/include/agent_registration.hpp
+++ b/src/agent/include/agent_registration.hpp
@@ -21,7 +21,7 @@ namespace agent_registration
                           std::string password,
                           const std::string& key,
                           const std::string& name,
-                          const std::string& configPath);
+                          std::optional<std::string> configPath);
         bool Register(http_client::IHttpClient& httpClient);
 
     private:
--- a/src/agent/include/agent_registration.hpp
+++ b/src/agent/include/agent_registration.hpp
@@ -17,7 +17,11 @@ namespace agent_registration
     class AgentRegistration
     {
     public:
-        AgentRegistration(std::string user, std::string password, const std::string& key, const std::string& name);
+        AgentRegistration(std::string user,
+                          std::string password,
+                          const std::string& key,
+                          const std::string& name,
+                          const std::string& configPath);
         bool Register(http_client::IHttpClient& httpClient);
 
     private:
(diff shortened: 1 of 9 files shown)


aritosteles (approved):
LGTM!

src/agent/src/process_options_unix.cpp
```diff
@@ -58,18 +58,20 @@ void StopAgent()
...
     std::cout << "Options:\n";
     std::cout << "     --run                Start wazuh-agent\n";
+    std::cout << "       [--config-path <path>]   Specify configuration file path (optional)\n";
     std::cout << "     --start              Start wazuh-agent daemon\n";
     std::cout << "     --status             Get wazuh-agent daemon status\n";
     std::cout << "     --stop               Stop wazuh-agent daemon\n";
     std::cout << "     --restart            Restart wazuh-agent daemon\n";
+    std::cout << "       [--config-path <path>]   Specify configuration file path (optional)\n";
     std::cout << "     --register-agent     Register wazuh-agent\n";
+    std::cout << "       [--config-path <path>]   Specify configuration file path (optional)\n";
```
jr0me, line 74: Should this be three times?

src/agent/include/agent_registration.hpp
```diff
@@ -17,7 +17,11 @@ namespace agent_registration
     class AgentRegistration
     {
     public:
-        AgentRegistration(std::string user, std::string password, const std::string& key, const std::string& name);
+        AgentRegistration(std::string user,
+                          std::string password,
+                          const std::string& key,
+                          const std::string& name,
+                          const std::string& configPath);
```
jr0me, line 24: It would be better if this is an `std::optional` otherwise we force some other uses to pass an empty string.

sdvendramini (approved):

jr0me (approved):

Issue #246, opened 2024-10-25
Let the agent's CLI key option be optional


vikman90:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

Currently, option `--key` is mandatory when trying to register the agent.

Let this parameter be optional. Should it not be defined, create a random one.

sdvendramini:
## 30/10/2024
I was resolving [PR](https://github.com/wazuh/wazuh-agent/pull/252) comments and adding tests and documentation.

PR #252, opened 2024-10-28
Enhancement/246 Let the agent's CLI key option be optional

sdvendramini:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/246|

## Description

With these changes, the key can now be automatically generated if not provided, making the `--key` argument optional during agent registration process. Also if the key is provided, is validated to ensure it is 32 alphanumeric characters.


src/agent/agent_info/include/agent_info.hpp
```diff
@@ -19,6 +19,9 @@ class AgentInfo
     void SetUUID(const std::string& uuid);
     void SetGroups(const std::vector<std::string>& groupList);
 
+    std::string CreateKey();
```
TomasTurina, line 22:
```suggestion
    std::string CreateKey() const;
```

TomasTurina, line 22: These two functions seem to be like utils, I don't know if this is the right place for them.

src/agent/src/agent_registration.cpp
```diff
@@ -23,7 +25,22 @@ namespace agent_registration
...
-        m_agentInfo.SetKey(key);
+        if (!key.empty())
+        {
+            if (m_agentInfo.ValidateKey(key))
+            {
+                m_agentInfo.SetKey(key);
+            }
+            else
+            {
+                std::cout << "--key argument must be alphanumeric and 32 characters in lengt\n";
```
TomasTurina, line 36:
```suggestion
                std::cout << "--key argument must be alphanumeric and 32 characters in length\n";
```

src/agent/agent_info/include/agent_info.hpp
TomasTurina: Maybe they can be private and be called inside `SetKey`.

src/agent/src/process_options.cpp
```diff
@@ -28,6 +28,6 @@ void RegisterAgent(const std::string& user,
     }
     else
     {
-        LogError("--user, --password and --key args are mandatory");
+        LogError("--user, --password args are mandatory");
```
TomasTurina, line 31:
```suggestion
        LogError("--user and --password args are mandatory");
```

src/agent/agent_info/src/agent_info.cpp
```diff
@@ -80,3 +83,33 @@ void AgentInfo::SetGroups(const std::vector<std::string>& groupList)
...
+    {
+        key += charset[distribution(generator)];
+    }
+
+    return key;
+}
+
+bool AgentInfo::ValidateKey(const std::string& key)
+{
+    static constexpr size_t key_length = 32;
```
TomasTurina, line 106: We have this two times, can we have just one globally for the class?

src/agent/agent_info/include/agent_info.hpp
sdvendramini: Fixed!

src/agent/src/agent_registration.cpp
sdvendramini: Fixed!

src/agent/src/process_options.cpp
sdvendramini: Fixed!

src/agent/agent_info/src/agent_info.cpp
sdvendramini: Now it's a `constexpr` within an anonymous namespace in the class implementation file.

vikman90 (changes requested):
@sdvendramini Please don't forget to add the tests corresponding to the new `CreateKey()` and `ValidateKey()` methods.

src/agent/agent_info/src/agent_info.cpp
```diff
@@ -80,3 +92,31 @@ void AgentInfo::SetGroups(const std::vector<std::string>& groupList)
     agentInfoPersistance.SetGroups(groupList);
     m_groups = groupList;
 }
+
+std::string AgentInfo::CreateKey()
+{
+    constexpr std::string_view charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+    boost::random::mt19937 generator {boost::random::random_device {}()};
+    boost::random::uniform_int_distribution<size_t> distribution(0, charset.size() - 1);
```
jr0me, line 101: Isn't this possible using `<random`> ?

src/agent/src/agent_registration.cpp
```diff
@@ -41,16 +45,12 @@ namespace agent_registration
 
         if (!token.has_value())
         {
-            LogError("Failed to authenticate with the manager");
+            std::cout << fmt::format("Failed to authenticate with the manager\n");
```
jr0me, line 48: This doesn't seem to need `fmt::format`.

src/agent/agent_info/src/agent_info.cpp
```diff
@@ -80,3 +92,31 @@ void AgentInfo::SetGroups(const std::vector<std::string>& groupList)
...
+bool AgentInfo::ValidateKey(const std::string& key)
+{
+    if (key.length() != KEY_LENGTH)
+    {
+        return false;
+    }
+
+    return std::ranges::all_of(key, ::isalnum);
+
+    return true;
```
jr0me, line 121: This line seems to be a leftover.

src/agent/agent_info/src/agent_info.cpp
```diff
@@ -80,3 +92,31 @@ void AgentInfo::SetGroups(const std::vector<std::string>& groupList)
...
+}
+
+bool AgentInfo::ValidateKey(const std::string& key)
+{
+    if (key.length() != KEY_LENGTH)
+    {
+        return false;
+    }
+
+    return std::ranges::all_of(key, ::isalnum);
```
jr0me, line 119: Could be simplified with

```
return key.length() == KEY_LENGTH && std::ranges::all_of(key, ::isalnum);
```

src/agent/src/agent_registration.cpp
```diff
@@ -1,8 +1,8 @@
 #include <agent_registration.hpp>
 
-#include <logger.hpp>
-
 #include <boost/beast/http.hpp>
+#include <fmt/format.h>
```
jr0me, line 4: It looks like if we are going to use fmt it should be included privately in the target include libraries of the Agent target. From what I gather, fmt is a public dependency of Logger... but it is actually not used at all by Logger... I may be wrong, but we could try removing fmt::format from Logger and if we actually need it here, add it privately to the Agent target.

src/agent/src/agent_registration.cpp
sdvendramini: fixed!

src/agent/agent_info/src/agent_info.cpp
sdvendramini: fixed!

src/agent/agent_info/src/agent_info.cpp
sdvendramini: Yes, it's true! Changed!

src/agent/src/agent_registration.cpp
sdvendramini: Yes, I was able to remove `fmt` from the `Logger` and include it as a private library in the `Agent` without problems! I also stopped using fmt for these simpler messages.

jr0me (approved):
LGTM

TomasTurina (approved):

vikman90 (approved):

TomasTurina (approved):

Issue #243, opened 2024-10-25
Let the agent accept a server address as URL


vikman90:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

Replace the `manager_ip` setting with `server_url`. This setting shall accept a URL:

- Accept both HTTP and HTTPS.
- Accept IP addresses, names, and FQDNs.

## Example

```yaml
server_url:       https://server.wazuh.com.test:27000
registration_url: https://server.wazuh.com.test:55000
```

vikman90:
## Issue Update

### October 31
- PR #258
- Applying requested changes.

aritosteles:
# Update

- Config settings renamed to server_url and registration_url. Port and UseHttps settings removed. All classes, functions and tests involved were refactored.
- Added boost::urls for url parsing and validation.
     - Parsing and validation done in HttpRequestParams, Communicator and AgentRegistration classes.
     - In case of ill-formated urls in config file a log entry is created, no crash or exceptions thrown. More thorough validation of config settings is in order. 
- HttpRequestParams class moved to its own hpp and cpp files.
-Added flag `--output-on-failure` to ctest execution to help debug failing tests.

PR is on final review

PR #258, opened 2024-10-28
Let the agent accept a server address as URL

aritosteles:
|Related issue|
|---|
|[243](https://github.com/wazuh/wazuh-agent/issues/243)|

## Description
Let the agent accept a server address as URL

  - [x] Linux

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -37,5 +37,5 @@ runs:
         run: |
           set +e
           cd build
-          ctest -C RelWithDebInfo -j $(nproc)
+          ctest -C RelWithDebInfo --output-on-failure -j $(nproc)
         shell: bash
--- a/etc/config/wazuh.conf
+++ b/etc/config/wazuh.conf
@@ -1,8 +1,6 @@
 [agent]
-server_mgmt_api_port = "55000"
-agent_comms_api_port = "27000"
-manager_ip = "localhost"
-https_enabled = "yes"
+server_url = "https://localhost:27000"
+registration_url = "https://localhost:55000"
 
 [inventory]
 enabled = true
--- a/src/agent/communicator/include/http_client.hpp
+++ b/src/agent/communicator/include/http_client.hpp
@@ -35,17 +35,13 @@ namespace http_client
         boost::beast::http::response<boost::beast::http::dynamic_body>
         PerformHttpRequest(const HttpRequestParams& params) override;
 
-        std::optional<std::string> AuthenticateWithUuidAndKey(const std::string& host,
-                                                              const std::string& port,
+        std::optional<std::string> AuthenticateWithUuidAndKey(const std::string& serverUrl,
                                                               const std::string& uuid,
-                                                              const std::string& key,
-                                                              const bool useHttps) override;
+                                                              const std::string& key) override;
 
-        std::optional<std::string> AuthenticateWithUserPassword(const std::string& host,
-                                                                const std::string& port,
+        std::optional<std::string> AuthenticateWithUserPassword(const std::string& serverUrl,
                                                                 const std::string& user,
-                                                                const std::string& password,
-                                                                const bool useHttps) override;
+                                                                const std::string& password) override;
 
         boost::beast::http::response<boost::beast::http::dynamic_body>
         PerformHttpRequestDownload(const HttpRequestParams& params, const std::string& dstFilePath) override;
--- a/src/agent/communicator/include/http_request_params.hpp
+++ b/src/agent/communicator/include/http_request_params.hpp
@@ -1,0 +1,29 @@
+#pragma once
+
+#include <boost/beast.hpp>
+
+#include <string>
+
+namespace http_client
+{
+    struct HttpRequestParams
+    {
+        boost::beast::http::verb Method;
+        std::string Host;
+        std::string Port;
+        std::string Endpoint;
+        bool Use_Https;
+        std::string Token;
+        std::string User_pass;
+        std::string Body;
+
+        HttpRequestParams(boost::beast::http::verb method,
+                          const std::string& serverUrl,
+                          std::string endpoint,
+                          std::string token = "",
+                          std::string userPass = "",
+                          std::string body = "");
+
+        bool operator==(const HttpRequestParams& other) const;
+    };
+} // namespace http_client
--- a/src/agent/communicator/include/ihttp_client.hpp
+++ b/src/agent/communicator/include/ihttp_client.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <http_request_params.hpp>
 #include <ihttp_socket.hpp>
 
 #include <boost/asio.hpp>
@@ -13,45 +14,6 @@
 
 namespace http_client
 {
-    struct HttpRequestParams
-    {
-        boost::beast::http::verb Method;
-        std::string Host;
-        std::string Port;
-        std::string Endpoint;
-        bool Use_Https;
-        std::string Token;
-        std::string User_pass;
-        std::string Body;
-
-        HttpRequestParams(boost::beast::http::verb method,
-                          const std::string& host,
-                          const std::string& endpoint,
-                          const std::string& token = "",
-                          const std::string& user_pass = "",
-                          const std::string& body = "")
-            : Method(method)
-            , Host(host)
-            , Endpoint(endpoint)
-            , Token(token)
-            , User_pass(user_pass)
-            , Body(body)
-        {
-            boost::urls::url_view url(host);
-
-            Use_Https = url.scheme().empty() || url.scheme() == "https";
-            Host = url.host();
-            Port = url.port();
-        }
-
-        bool operator==(const HttpRequestParams& other) const
-        {
-            return Method == other.Method && Host == other.Host && Port == other.Port && Endpoint == other.Endpoint &&
-                   Token == other.Token && User_pass == other.User_pass && Body == other.Body &&
-                   Use_Https == other.Use_Https;
-        }
-    };
-
     class IHttpClient
     {
     public:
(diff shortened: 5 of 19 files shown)


etc/config/wazuh.conf
```diff
@@ -1,7 +1,7 @@
 [agent]
 server_mgmt_api_port = "55000"
 agent_comms_api_port = "27000"
-manager_ip = "localhost"
+server_address = "localhost"
```
TomasTurina, line 4:
```suggestion
server_url = "localhost"
```
Let's use this name for now and also `registration_url` for the other API.

aritosteles: Done

src/agent/communicator/include/ihttp_client.hpp
```diff
@@ -25,21 +26,22 @@ namespace http_client
...
             : Method(method)
             , Host(host)
-            , Port(port)
             , Endpoint(endpoint)
-            , Use_Https(use_https)
             , Token(token)
             , User_pass(user_pass)
             , Body(body)
         {
+            boost::urls::url_view url(host);
```
TomasTurina, line 40: What if it is an invalid url?

src/agent/communicator/include/ihttp_client.hpp
```diff
@@ -25,21 +26,22 @@ namespace http_client
...
         HttpRequestParams(boost::beast::http::verb method,
                           const std::string& host,
-                          const std::string& port,
                           const std::string& endpoint,
-                          const bool use_https,
                           const std::string& token = "",
                           const std::string& user_pass = "",
                           const std::string& body = "")
             : Method(method)
             , Host(host)
```
TomasTurina, line 34: Remove.

src/agent/communicator/src/http_client.cpp
```diff
@@ -231,13 +226,18 @@ namespace http_client
...
         const auto reqParams = http_client::HttpRequestParams(
-            boost::beast::http::verb::post, host, port, "/security/user/authenticate", useHttps, "", basicAuth);
+            boost::beast::http::verb::post, host, "/security/user/authenticate", "", basicAuth);
 
         const auto res = PerformHttpRequest(reqParams);
 
         if (res.result() != boost::beast::http::status::ok)
         {
             LogError("Error: {}.", res.result_int());
+            LogError("Host: {}, Port: {}, Endpoint: {}, Https: {}",
```
TomasTurina, line 236: Why do you need this log? Why it is only here?

src/vcpkg.json
```diff
@@ -1,123 +1,129 @@
...
+      }
+    ],
+    "overlay-ports": [
+      "./ports-overlay/procps",
+      "./ports-overlay/libdb",
+      "./ports-overlay/librpm"
+    ]
   }
-}
+}
```
TomasTurina, line 129: Missing empty line at the end.

src/agent/communicator/include/http_client.hpp
```diff
@@ -35,17 +35,12 @@ namespace http_client
         boost::beast::http::response<boost::beast::http::dynamic_body>
         PerformHttpRequest(const HttpRequestParams& params) override;
 
-        std::optional<std::string> AuthenticateWithUuidAndKey(const std::string& host,
-                                                              const std::string& port,
-                                                              const std::string& uuid,
-                                                              const std::string& key,
-                                                              const bool useHttps) override;
+        std::optional<std::string>
+        AuthenticateWithUuidAndKey(const std::string& host, const std::string& uuid, const std::string& key) override;
```
jr0me, line 39: Can we change the `host` name on the first argument? Should be address or url now...

src/agent/communicator/src/http_client.cpp
aritosteles: PerformHttpRequest is failing on non https requests. I needed to check what the parameters were. I actually expected to have some feedback from you on whether this log was ok to keep. Having just the result is not that useful.

src/agent/communicator/include/ihttp_client.hpp
```diff
@@ -25,21 +26,22 @@ namespace http_client
...
-            , Use_Https(use_https)
             , Token(token)
             , User_pass(user_pass)
             , Body(body)
         {
+            boost::urls::url_view url(host);
+
+            Use_Https = url.scheme().empty() || url.scheme() == "https" ? true : false;
+            Host = url.host();
+            Port = url.port();
```
jr0me, line 44: Since the linking on the cmake file is private so we should use it privately by moving the constructor definition to the source file (together with the include).

src/agent/communicator/include/ihttp_client.hpp
```diff
@@ -25,21 +26,22 @@ namespace http_client
...
-            , Port(port)
             , Endpoint(endpoint)
-            , Use_Https(use_https)
             , Token(token)
             , User_pass(user_pass)
             , Body(body)
         {
+            boost::urls::url_view url(host);
+
+            Use_Https = url.scheme().empty() || url.scheme() == "https" ? true : false;
```
jr0me, line 42: this should be

```
Use_Https = url.scheme().empty() || url.scheme() == "https";

File: src/agent/communicator/include/ihttp_client.hpp [hunk 9]
```diff
@@ -69,17 +71,11 @@ namespace http_client
         virtual boost::beast::http::response<boost::beast::http::dynamic_body>
         PerformHttpRequest(const HttpRequestParams& params) = 0;
 
-        virtual std::optional<std::string> AuthenticateWithUuidAndKey(const std::string& host,
-                                                                      const std::string& port,
-                                                                      const std::string& uuid,
-                                                                      const std::string& key,
-                                                                      const bool useHttps) = 0;
+        virtual std::optional<std::string>
+        AuthenticateWithUuidAndKey(const std::string& host, const std::string& uuid, const std::string& key) = 0;
```
jr0me, line 75: Same change on `host` argument name.

src/vcpkg.json
jr0me, line 1: We shouldn't be changing the formating imo. I understand that this was an automatic linting, but since we don't have a unified linter for json files, (like we do have with c++ and clang format) this can potentially change with each developer's environment.

src/agent/communicator/include/ihttp_client.hpp
aritosteles: Right! Done

src/agent/communicator/include/ihttp_client.hpp
aritosteles: Done

src/agent/communicator/include/http_client.hpp
aritosteles: Done

src/agent/communicator/include/ihttp_client.hpp
aritosteles: Done

src/vcpkg.json
aritosteles: I reverted the file to the version in master and re applied my changes. The format mods were made without me noticing, thanks for the heads up!

src/vcpkg.json
aritosteles: Fixed

etc/config/wazuh.conf
```diff
@@ -1,8 +1,6 @@
 [agent]
-server_mgmt_api_port = "55000"
-agent_comms_api_port = "27000"
-manager_ip = "localhost"
-https_enabled = "yes"
+server_url = https://localhost:27000
+registration_url = https://localhost:55000
```
Nicogp, line 3: the value of `server_url `and `registration_url `must be enclosed in double quotation marks:

```
server_url = "https://localhost:27000"
registration_url = "https://localhost:55000"
```

aritosteles: Fixed!

src/agent/communicator/include/ihttp_client.hpp
aritosteles: Done

src/agent/communicator/src/http_client.cpp
TomasTurina: I would say to remove it. We can analyze what logs can we improve later.

src/agent/communicator/include/http_client.hpp
TomasTurina: This repeats in many other functions.

src/agent/communicator/src/http_client.cpp
aritosteles: Removed

src/agent/communicator/include/ihttp_client.hpp
aritosteles: This case has been handled so there is no crash or exception thrown. A more thorough validation of configuration settings is necessary, though.

src/agent/communicator/include/http_client.hpp
aritosteles: I changed it in HttpRequestParams constructor as well. I may be missing other occurrences.

.github/actions/compile/action.yml
```diff
@@ -37,5 +37,5 @@ runs:
         run: |
           set +e
           cd build
-          ctest -C RelWithDebInfo -j $(nproc)
+          ctest -C RelWithDebInfo --output-on-failure -j $(nproc)
```
jr0me, line 40: :ok_hand:

src/agent/communicator/include/http_request_params.hpp
```diff
@@ -0,0 +1,31 @@
+#pragma once
+
+#include <boost/asio.hpp>
+#include <boost/beast.hpp>
+#include <boost/url.hpp>
```
jr0me, line 5: Asio is not needed by this class, and Url can be moved to the source file.

jr0me (approved):
LGTM

Nicogp (approved):
LGTM

TomasTurina (approved):

Issue #251, opened 2024-10-25
Convert DBsync and Sysinfo from shared to static library


cborla:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

## **Description**

The objective of this issue is to remove shared libraries from the Inventory module. Specifically, the DBsync and Sysinfo libraries should be converted from shared to static to improve compatibility across platforms.

## **Tasks**
- [ ] Convert the DBsync library from shared to static.
- [ ] Convert the Sysinfo library from shared to static.
- [ ] Validate Inventory functionality across the following platforms:
  - [ ] Linux
  - [ ] Windows
  - [ ] macOS
- [ ] Enable the Module Manager test for Windows, which previously failed due to missing DLL files.

## **Acceptance Criteria**
- DBsync and Sysinfo libraries are successfully converted to static libraries without impacting existing functionality.
- Inventory module passes functionality tests on Linux, Windows, and macOS platforms.
- Windows Module Manager tests run successfully without DLL or dependency errors.
- No shared library dependencies remain in the Inventory module after conversion.

nbertoldo:
### Update 2024/10/25

- Converting _sysinfo_ and _dbsync_ from shared to static library.

nbertoldo:
### Update 2024/10/30

When changing _sysinfo_ from `SHARED` to `STATIC` library I get these errors that occur at the time of linking _wazuh-agent_ executable:

```console
[ 95%] Linking CXX executable wazuh-agent
/usr/bin/ld: vcpkg_installed/x64-linux/debug/lib/librpm.a(rpmscript.o): in function `next_file':
/home/vagrant/wazuh-agent/src/vcpkg/buildtrees/librpm/x64-linux-dbg/lib/.././../src/9ed556c157-1b82a7a6d4.clean/lib/rpmscript.c:101:(.text+0x83): undefined reference to `lua_touserdata'
/usr/bin/ld: /home/vagrant/wazuh-agent/src/vcpkg/buildtrees/librpm/x64-linux-dbg/lib/.././../src/9ed556c157-1b82a7a6d4.clean/lib/rpmscript.c:102:(.text+0xad): undefined reference to `lua_pushstring'
...
```

It appears that librpm cannot find references to lua functions.

PR #262, opened 2024-10-29
Convert DBsync and Sysinfo from shared to static library

nbertoldo:
|Related issue|
|---|
| #251 |

## Description

This PR removes shared libraries from the Inventory module. Specifically, the `DBsync` and `Sysinfo` libraries should be converted from shared to static to improve platform compatibility.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X


cborla (approved):
LGTM

src/modules/inventory/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@ project(Inventory)
 include(../../cmake/CommonSettings.cmake)
 set_common_settings()
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /WX-")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /WX-")
+    add_compile_options(/w)
+endif()
```
jr0me, line 12: I wonder if it could work like this...

```
if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
    target_compile_options(Inventory PRIVATE /WX- /w)
endif()
```

(should probably come after line 23 (because of the `add_library`...)

If it still doesn't work we could try to find the targets that need these flags, instead of applying them to the entire project.

src/modules/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ message(STATUS "-------------------")
 
 if(ENABLE_INVENTORY)
     add_subdirectory(inventory)
-    target_link_libraries(ModuleManager PRIVATE Inventory)
+    target_link_libraries(ModuleManager PUBLIC Inventory)
```
jr0me, line 20: It doesn't work if it is `PRIVATE`?

jr0me (approved):
LGTM

A couple of notes:
* The two commits I added could be squashed, and their titles changed, I was testing at the moment so I didn't adhere to the convention of marking commits with fix, feat, etc...
* In some of these cmake files there are things to review later on; ie: one is building with c++14, other with c++17, and they set CXX flags that could potentially overwrite the flags of the parent scope (we should use the configureTarget cmake function or use target_compile_options) but it is beyond the scope of this issue.

src/modules/inventory/CMakeLists.txt
nbertoldo: Done!

PR #264, opened 2024-10-31
Fix crash hazard in Inventory module

vikman90:
|Related issue|
|---|
|https://github.com/wazuh/wazuh/issues/26488|

This is a port of PR:
- https://github.com/wazuh/wazuh/pull/26647

It aims to fix the following bugs that may lead the agent to crash:

1. The hashing functions, which rely on OpenSSL (Crypto), delegate automatic initialization. This initialization installs a cleanup trigger for program shutdown. When this cleanup is executed, some threads may still be using OpenSSL, and this causes a segmentation fault.
   **Fix:** Manually initialize OpenSSL, and disable the cleanup trigger.
2. The DataProvider component (used by Inventory) has an uninitialized buffer, which can lead to buffer overrun.
   **Fix:** Zero-initialize the buffer.


cborla (approved):
LGTM

TomasTurina (approved):

Issue #103, opened 2024-08-22
Change ConfigurationParser to parse YAML instead of TOML


TomasTurina:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

## Description

The current implementation of the `ConfigurationParser` can parse TOML files, but after some consideration we need to change this and make the parser able to interpret YAML files, as that will be the configuration format across all Wazuh components.

Tasks to be accomplished:

- [x] Select and use a common and maintained library, mention the advantages and disadvantages of its use.
- [x] Replace the TOML library with the selected YAML library.
- [x] Rename the default configuration file to _wazuh-agent.yml_, which must be installed into _/etc/wazuh-agent/wazuh-agent.yml_.
- [x] Adapt and create new UT for the 'ConfigurationParser'.

Nicogp:
## Research of yaml libraries available for C++

The official [YAML page](https://yaml.org/) lists the following libraries:

> YAML Frameworks and Tools:
>   C/C++:
>   - [libfyaml](https://github.com/pantoniou/libfyaml)      # "C" YAML 1.2 processor ([YTS](https://yaml.org/#yts))
>   - [libyaml](http://pyyaml.org/wiki/LibYAML)       # "C" Fast YAML 1.1 ([YTS](https://yaml.org/#yts))
>   - [libcyaml](https://github.com/tlsa/libcyaml)      # YAML de/serialization of C data (using libyaml)
>   - [yaml-cpp](https://github.com/jbeder/yaml-cpp/)      # C++ YAML 1.2 implementation
the only one native for C++ is yaml-cpp, the rest are native for C.

If we take [cppreference ](https://en.cppreference.com/w/cpp/links/libs)as a source, here it names only yaml-cpp:
![image](https://github.com/user-attachments/assets/fe6b0fd1-32ea-4acd-8a68-5f08ecbdde4b)

### Comparison of libraries: 
While researching, we also found the rapidyaml library that was included for comparison.

**YAML-CPP:**
https://github.com/jbeder/yaml-cpp

- Advantages:

  - Supports recent YAML versions. Full support for YAML specification
  - Open source, widely used and supported
  - Easy to install (available in vcpkg).
  - Provides intuitive data access for hierarchical structures.
  - Integrates smoothly into projects using CMake.

- Disadvantages:

  - Can be heavy for small projects or those with strict performance requirements.

**libyaml:**
https://pyyaml.org/wiki/LibYAML

- Advantages:

[...]

PR #266, opened 2024-11-02
Changes in ConfigurationParser module to parse yaml instead of toml

Nicogp:
|Related issue|
|---|
|#103|

## Description

This PR adds changes to the ConfigurationParser module so that it now parses yaml files instead of toml files.

- Includes the renaming of the configuration file to `wazuh-agent.yml`
- Includes test fixes for modules that use the configurationParser module.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X


src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -14,20 +14,32 @@ namespace configuration
...
+                (
+                    [&current](const auto& key)
+                    {
+                        current = current[key];
+                        if (!current.IsDefined())
+                        {
+                            throw YAML::Exception(YAML::Mark::null_mark(), "Key not found: " + std::string(key));
+                        }
+                    }(keys),
+                    ...);
```
jr0me, line 67: Is it possible to format it like:

```
(
    [&current](const auto& key)
    {
        current = current[key];
        if (!current.IsDefined())
        {
            throw YAML::Exception(YAML::Mark::null_mark(), "Key not found: " + std::string(key));
        }
    }(keys),
    ...
);
```

Or does clang format get in the way?

src/agent/configuration_parser/tests/configuration_parser_test.cpp
```diff
@@ -2,11 +2,81 @@
...
+#include <fstream>
 #include <map>
 #include <memory>
 
 using namespace configuration;
 
+class ConfigurationParserFileTest : public ::testing::Test
+{
+protected:
+    std::filesystem::path tempConfigFilePath;
```
jr0me, line 15: Could it be `m_tempConfigFilePath`?

src/agent/configuration_parser/tests/configuration_parser_test.cpp
```diff
@@ -2,11 +2,81 @@
...
+                - /var/log/other.log
+                reload_interval: 120
+                file_wait: 1000
+        )";
+        outFile.close();
+    }
+
+    void TearDown() override
+    {
+        std::filesystem::remove(tempConfigFilePath);
```
jr0me, line 42: How necessary is this remove? I ask before if this crashes somehow the file won't be cleaned up.

src/modules/logcollector/src/logcollector.cpp
```diff
@@ -37,13 +37,13 @@ void Logcollector::SetupFileReader(const configuration::ConfigurationParser& con
     long reloadInterval = config::DEFAULT_RELOAD_INTERVAL;
 
     try {
-        fileWait = m_enabled = configurationParser.GetConfig<long>("logcollector", "file_wait");
+        fileWait = configurationParser.GetConfig<long>("logcollector", "file_wait");
```
jr0me, line 40: Nice catch @Nicogp 

I'm negatively surprised that this wasn't catch any warning or static analysis. I played around in godbolt ( https://godbolt.org/z/jb8564oz5 ) and found that there's one warning we should have seen:

```
[warning: Although the value stored to 'enabled' is used in the enclosing expression, the value is never actually read from 'enabled' [clang-analyzer-deadcode.DeadStores]](javascript:;)
```

Something to look into later on...

src/agent/configuration_parser/include/configuration_parser.hpp
Nicogp: clang_format reorder it as it was originally

src/agent/configuration_parser/tests/configuration_parser_test.cpp
Nicogp: If the failure is due to an `EXPECT_*` the TearDown would be executed, the `EXPECT_*` allows the test to continue to the end, registering any failure, but without interrupting the execution. That is why I decided to use `EXPECT_*` instead of `ASSERT_*`.
However, I could add some more controls in case there is an exception for some other reason.
In the tests:
```
    try
    {
        const auto parser = std::make_unique<configuration::ConfigurationParser>(m_tempConfigFilePath);
    }
    catch(const std::exception& e)
    {
        std::filesystem::remove(m_tempConfigFilePath);
        throw;
    }
```
What do you think?

jr0me: Right, the issue would come up if the application crashes before executing the `TearDown`. The problem with this try catch is that it would remove the file, but wouldn't create the parser. I think that checking that there's no file, and removing them if there are any, in the constructor, would be enough.

Nicogp: by default, ofstream overwrites the file in case it exists.

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -14,20 +14,32 @@ namespace configuration
     class ConfigurationParser
     {
     private:
-        toml::value tbl;
+        YAML::Node config;
```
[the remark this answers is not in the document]
Nicogp: Done!

src/agent/configuration_parser/tests/configuration_parser_test.cpp
Nicogp: Done!

TomasTurina (approved):

Issue #318, opened 2024-11-19
Agent Registration Fails Due to 201 Status Code


vikman90:
**Parent Issue:** #241 

## Description

When an agent successfully registers, the API responds with status code `201`. However, the agent incorrectly interprets this status as an error:
```
Registration error: 201.
wazuh-agent registration failed
```  

## Steps to Reproduce

1. Register an agent against a properly configured API.
2. Observe the registration process logs.
3. Note that despite a `201` response, the agent reports the registration as failed.

## Expected Behavior

The agent should treat `201` as a success code for registration, not as an error.

## Proposal

Update the agent to correctly handle a `201` status code as a successful registration response.

Issue #314, opened 2024-11-19
Agent Saves Registration Information Even on Failure


vikman90:
**Parent Issue:** #241

## Description
When attempting to register an agent and the process fails with errors such as:
```
Failed to authenticate with the manager
wazuh-agent registration failed
```
The agent still creates and populates the _agent_info.db_ file:
```
select * from agent_info;
dummy|ku3iIzPRNJjwIXKNnTU3qFv6QhodAu8k|95635cae-0c0d-486e-94a5-0fe29dd6972f
```

## Steps to Reproduce
1. Attempt to register the agent with invalid credentials.
2. Observe the error messages in the logs.
3. Check the contents of the _agent_info.db_ file.

## Expected Behavior
If the registration process fails, the database should not be altered.

Issue #281, opened 2024-11-07
Test Wazuh Agent with Mock Server Using Docker and Imposter


vikman90:
## Description
To thoroughly test the Wazuh Agent’s interaction with a server, we need to set up a mock server capable of handling several key use cases. Specifically, this includes:

1. **Agent Registration** – Testing the process of registering a new agent with the server.
2. **Stateless Message Sending** – Verifying the agent’s ability to send messages without expecting a response or maintaining a session.
3. **Stateful Message Sending** – Testing the agent’s capacity to send messages while maintaining state (e.g., sessions).

To implement this, we will create a Docker container using [Imposter](https://imposter.sh) as the mock server, which will facilitate these tests in a controlled environment.

## Requirements
- The Wazuh Agent requires SSL to communicate with the server, so SSL support must be enabled in the mock server configuration.
- The mock server should be configured to handle both stateless and stateful interactions as per the use cases above.

## Tasks
1. Set up a Docker environment with Imposter as the mock server.
2. Configure Imposter to support SSL connections.
3. Define mock endpoints for:
   - Agent registration
   - Stateless message handling
   - Stateful message handling
4. Document the setup and configuration steps to replicate the testing environment.
5. Implement test cases to verify that the Wazuh Agent can successfully communicate with the mock server in each of the defined scenarios.

PR #319, opened 2024-11-19
Fix Agent Registration Behavior

vikman90:
## Summary  

This pull request addresses two issues:  
- #314
- #318

## Changes  

- **Database Handling:**
   - Refactored the `AgentInfo` component to delay database writes until registration is confirmed.
    - Introduced a `Save()` method to persist data only after successful registration.
- **Registration Status Code:**
   Modified the expected status code for registration responses from `200` to `201`.

## Tests

- [X] **AgentInfo Tests:**
   - Updated existing tests to validate the new behavior of `Set()` and `Save()`.
   - Verified that database writes only occur after `Save()` is called.
- [X] **AgentRegistration Tests:**
   - Adapted tests to account for the `201` status code as a success indicator.
   - Ensured the agent correctly interprets `201` and does not fail the registration process.
- [X] **Manual Testing:**
   - Tested with a mock server using Imposter: #281
     - Verified that `201` responses result in successful registration.
     - Confirmed that the database remains unmodified in cases of registration failure.

> [!IMPORTANT] 
> In this change, we simplify the `RegistrationTestSuccessWithEmptyKey` test, since this test makes the agent produce a random key. Since persistence is removed, the mock does not have access to the new key. However, the persistence behavior of `AgentRegistration` is checked in `RegistrationSuccess`.

#### Registration failure

```shell
./wazuh-agent --config-file wazuh-agent.yml --register-agent --user wazuh --password topsecret --name other
```

> Failed to authenticate with the manager
> wazuh-agent registration failed

```shell
sqlite3 agent_info.db 'select * from agent_info'
```

> ||

#### Registration success

```shell
./wazuh-agent --config-file wazuh-agent.yml --register-agent --user wazuh --password topsecret --name dummy
```

> wazuh-agent registered

```shell
sqlite3 agent_info.db 'select * from agent_info'
```

> dummy|vEcz2euGDR2mTAxn4Fu6KrCIOhmsfq7U|114563d6-a9ae-4aec-aa58-7f0ee0500bab

### DIFF
--- a/src/agent/tests/agent_registration_test.cpp
+++ b/src/agent/tests/agent_registration_test.cpp
@@ -71,11 +71,16 @@ TEST_F(RegisterTest, RegistrationTestSuccess)
     AgentInfoPersistance agentInfoPersistance;
     agentInfoPersistance.ResetToDefault();
 
-    registration = std::make_unique<agent_registration::AgentRegistration>(
-        "user", "password", "4GhT7uFm1zQa9c2Vb7Lk8pYsX0WqZrNj", "agent_name", std::nullopt);
     SysInfo sysInfo;
     agent = std::make_unique<AgentInfo>([&sysInfo]() mutable { return sysInfo.os(); },
                                         [&sysInfo]() mutable { return sysInfo.networks(); });
+
+    agent->SetKey("4GhT7uFm1zQa9c2Vb7Lk8pYsX0WqZrNj");
+    agent->SetName("agent_name");
+    agent->Save();
+
+    registration = std::make_unique<agent_registration::AgentRegistration>(
+        "user", "password", "4GhT7uFm1zQa9c2Vb7Lk8pYsX0WqZrNj", "agent_name", std::nullopt);
 
     MockHttpClient mockHttpClient;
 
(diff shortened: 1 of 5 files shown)


src/agent/tests/agent_registration_test.cpp
```diff
@@ -152,29 +157,16 @@ TEST_F(RegisterTest, RegistrationTestSuccessWithEmptyKey)
...
-                                             agent->GetHeaderInfo(),
-                                             "token",
-                                             "",
-                                             bodyJson);
-
     boost::beast::http::response<boost::beast::http::dynamic_body> expectedResponse;
     expectedResponse.result(boost::beast::http::status::ok);
 
-    EXPECT_CALL(mockHttpClient, PerformHttpRequest(testing::Eq(reqParams))).WillOnce(testing::Return(expectedResponse));
+    EXPECT_CALL(mockHttpClient, PerformHttpRequest(testing::_)).WillOnce(testing::Return(expectedResponse));
```
jr0me, line 169: The previous version had a specific expectation on what parameters it would use to perform an http request.

TomasTurina (approved):

Issue #286, opened 2024-11-11
Standardize Time Unit in Agent Configuration Parser


vikman90:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

## Description

Currently, the agent's configuration includes time-based options, each specified with implicit or explicit time units. For example:

```yaml
agent:
  retry_interval_secs: 30
inventory:
  interval: 3600
logcollector:
  reload_interval: 60
  file_wait: 500
```

Here:
- `retry_interval_secs` explicitly indicates seconds with the `_secs` suffix.
- `interval`, `reload_interval` and `file_wait` assume seconds but do not specify a unit.

To improve consistency and user experience, I propose the following changes:

1. **Adopt Seconds as the Default Time Unit**  
   All time-based options will default to seconds, making seconds the standard across the configuration.
   
2. **Rename `retry_interval_secs` to `retry_interval`**  
   By removing the `_secs` suffix, we align with the naming convention of other time-based options, reinforcing the default unit as seconds.

3. **Implement Support for Standard Time Suffixes**  
   Add support for standard time suffixes to improve flexibility and readability. This enhancement will allow users to specify units directly, supporting:
   - **`ms` for milliseconds**
   - **`s` for seconds**
   - **`m` for minutes**
   - **`h` for hours**
   - **`d` for days**

   For instance:
   ```yaml
   agent:
     retry_interval: 30s
   inventory:
     interval: 1h
   logcollector:
     reload_interval: 1m
     file_wait: 500ms
   ```

This change will enhance readability and maintain consistency across the configuration.

TomasTurina:
## Update

Working on the function that will validate and parse time units.

TomasTurina:
## Update

The code is almost ready, missing some tests to finish.

PR #324, opened 2024-11-20
Configuration parser: Standardize Time Units

TomasTurina:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/286|

## Description

This PR introduces a new way to read time value configurations in the configuration parser component:

- Time suffixes:
  - ms: milliseconds
  - s: seconds
  - m: minutes
  - h: hours
  - d: days
- Seconds by default.
- All time units are internally converted to milliseconds. They are stored in `std::time_t` variables.
- Adapt all modules to sleep using milliseconds.
- Some fixes in the mock server.

## Configuration options

```yml
agent:
  retry_interval: 30s
inventory:
  interval: 1h
logcollector:
  reload_interval: 1m
  file_wait: 500ms
```

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X


src/agent/configuration_parser/src/configuration_parser.cpp
```diff
@@ -71,4 +76,47 @@ namespace configuration
...
+            multiplier = A_DAY_IN_MILLIS;
+        }
+        else
+        {
+            // By default, assume seconds
+            number = option;
+            multiplier = A_SECOND_IN_MILLIS;
+        }
+
+        if (all_of(number.begin(), number.end(), ::isdigit) == false)
```
jr0me, line 115: `std::all_of` and `std::isdigit` should be used. Also, this needs the `<algorithm> header and ... <cctype> for isdigit?

src/agent/configuration_parser/src/configuration_parser.cpp
```diff
@@ -71,4 +76,47 @@ namespace configuration
...
+        {
+            throw std::invalid_argument("Invalid time unit: " + option);
+        }
+
+        return static_cast<unsigned long>(std::stoul(number) * multiplier);
```
jr0me: Should there be a maximum value?

src/agent/configuration_parser/include/configuration_parser.hpp
```diff
@@ -75,5 +75,15 @@ namespace configuration
...
+
+        /// @brief Converts a time unit represented as a string to an unsigned long value (ms).
+        /// @param option A string representing a time unit.
+        /// @return The corresponding unsigned long value.
+        /// @throws std::invalid_argument if the string does not represent a valid time unit.
+        /// @details This function parses a string representing a time unit and returns the equivalent unsigned long
+        /// value. The time unit can be expressed in milliseconds (e.g. "1ms"), seconds (e.g. "1s"), minutes (e.g.
+        /// "1m"), hours (e.g. "1h"), or days (e.g. "1d"). If no unit is specified, the value is assumed to be in
+        /// seconds.
+        unsigned long ParseTimeUnit(const std::string& option);
```
jr0me: Can this function be private, const, and maybe static?

src/agent/communicator/include/communicator.hpp
```diff
@@ -2,9 +2,13 @@
 
 #include <ihttp_client.hpp>
 
+#include <config.h>
+#include <logger.hpp>
```
jr0me, line 6: Missing ctime for m_retryInterval

src/agent/communicator/include/communicator.hpp
```diff
@@ -2,9 +2,13 @@
...
+
 #include <boost/asio/awaitable.hpp>
 #include <boost/asio/steady_timer.hpp>
 #include <boost/beast/http/status.hpp>
+#include <boost/url.hpp>
```
jr0me, line 11: Looks like boost::url and logger now are dependencies with public linkage, the cmake file needs to be modified.

src/agent/src/agent.cpp
```diff
@@ -21,8 +21,8 @@ Agent::Agent(const std::string& configFile, std::unique_ptr<ISignalHandler> sign
           m_agentInfo.GetUUID(),
           m_agentInfo.GetKey(),
           [this]() { return m_agentInfo.GetHeaderInfo(); },
-          [this](std::string table, std::string key) -> std::optional<std::string>
-          { return m_configurationParser.GetConfig<std::string>(std::move(table), std::move(key)); })
+          [this]<typename T>(std::string table, std::string key) -> std::optional<T>
```
jr0me, line 24: :magic_wand: :mage_man:

src/agent/communicator/tests/communicator_test.cpp
```diff
@@ -34,11 +34,20 @@ namespace
             .set_payload_claim("exp", jwt::basic_claim<jwt::traits::nlohmann_json>(exp))
             .sign(jwt::algorithm::hs256 {"secret"});
     }
+
+    const auto FUNC = []<typename T>([[maybe_unused]] const std::string&,
```
jr0me, line 38: Couldn't this just return T{}? instead of trying to return a time_t? is it needed for these tests? Also, the name be changed to GetConfigFunc or something like that?

Is it needed at all? It was nullptr before.

TomasTurina: Yes, we need to define this function as it can't be nullptr.

src/agent/configuration_parser/src/configuration_parser.cpp
TomasTurina: I've tested this with very big numbers and there wasn't a problem.

aritosteles (approved):

jr0me (approved):
LGTM

PR #325, opened 2024-11-20
Add new option to set the data folder

### LINKED ISSUE #315: Introduce Data Directory Configuration
**Parent Issue:** #241

## Description
Currently, the agent creates its local databases in the current working directory.

## Proposal
1. Add a new configuration option, `path.data`, in the `wazuh-agent.yml` file.
2. Default value for `path.data`: `/var/lib/wazuh-agent`.
3. Store all local databases and modifiable files in the specified data directory.

cborla:
|Related issue|
|---|
|Closes #315|

## Description

This pull request introduces a new configuration parameter `data.path` to the Wazuh Agent's configuration file. This parameter specifies the directory where the agent stores variable and persistent data. 

By default, the agent will use platform-specific paths if the parameter is not explicitly set in the configuration file.

### Default Paths

- **Linux**: `/var/lib/wazuh-agent`
- **macOS**: `/Library/Application Support/wazuh-agent`
- **Windows**: `C:\ProgramData\wazuh-agent\lib`

### Behavior

- If `data.path` is not present in the configuration file, the agent uses the platform-specific default path.
- If the parameter is defined, the agent uses the user-specified directory.

The configuration file (`wazuh-agent.yml`) does not include this option by default, allowing the agent to apply the appropriate platform default.

## Tests Performed

- [x] **Default Directory**: Confirmed that the agent correctly uses the default path when `data.path` is not set.
- [x] **Custom Directory**: Verified that a user-defined path in `wazuh-agent.yml` is applied successfully.
- [x] **Migration Check**: Ensured that the existing databases are no longer stored in the previous directory.
- [x] **Unit Tests**: Updated and executed unit tests to validate the changes.


src/agent/multitype_queue/include/multitype_queue.hpp
```diff
@@ -44,10 +44,11 @@ class MultiTypeQueue : public IMultiTypeQueue
     /**
      * @brief Constructor.
      *
+     * @param dbFolderPath The path to the database folder.
      * @param size The maximum number of items in the queue.
      * @param timeout The timeout period in seconds.
      */
-    MultiTypeQueue(size_t size = DEFAULT_MAX, int timeout = DEFAULT_TIMEOUT_S);
+    MultiTypeQueue(const std::string& dbFolderPath, size_t size = DEFAULT_MAX, int timeout = DEFAULT_TIMEOUT_S);
```
TomasTurina, line 51: Don't we need a default value for the DB folder path?

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -8,13 +8,16 @@ namespace
 {
     const std::string AGENT_INFO_TABLE_NAME = "agent_info";
     const std::string AGENT_GROUP_TABLE_NAME = "agent_group";
+    const std::string AGENT_INFO_DB_NAME = "agent_info.db";
 } // namespace
 
-AgentInfoPersistance::AgentInfoPersistance(const std::string& dbPath)
+AgentInfoPersistance::AgentInfoPersistance(const std::string& dbFolderPath)
 {
+    auto dbFilePath = dbFolderPath + "/" + AGENT_INFO_DB_NAME;
```
jr0me, line 16: Can be const

src/agent/multitype_queue/src/multitype_queue.cpp
```diff
@@ -8,14 +8,16 @@
 
 constexpr int DEFAULT_TIMER_IN_MS = 100;
 
-MultiTypeQueue::MultiTypeQueue(size_t size, int timeout)
+MultiTypeQueue::MultiTypeQueue(const std::string& dbFolderPath, size_t size, int timeout)
     : m_maxItems(size)
     , m_timeout(timeout)
 {
+    auto dbFilePath = dbFolderPath + "/" + QUEUE_DB_NAME;
```
jr0me, line 15: Can be const

src/agent/src/agent.cpp
```diff
@@ -11,11 +12,13 @@
 #include <thread>
 
 Agent::Agent(const std::string& configFile, std::unique_ptr<ISignalHandler> signalHandler)
-    : m_agentInfo([this]() { return m_sysInfo.os(); }, [this]() { return m_sysInfo.networks(); })
-    , m_messageQueue(std::make_shared<MultiTypeQueue>())
-    , m_signalHandler(std::move(signalHandler))
-    , m_configurationParser(configFile.empty() ? configuration::ConfigurationParser()
+    : m_configurationParser(configFile.empty() ? configuration::ConfigurationParser()
                                                : configuration::ConfigurationParser(std::filesystem::path(configFile)))
+    , m_dataPath(m_configurationParser.GetConfig<std::string>("agent", "path.data").value_or(config::DEFAULT_DATA_PATH))
```
TomasTurina, line 17: Don't you see necessary to add `path.data` to the default configuration file?
(`etc/config/wazuh-agent.yml`)

TomasTurina (changes requested):
What about the `command_store.db`?

src/agent/tests/agent_test.cpp
```diff
@@ -1,33 +1,74 @@
...
 
 class MockSignalHandler : public ISignalHandler
 {
 public:
     MOCK_METHOD(void, WaitForSignal, (), (override));
 };
 
+void CreateTempConfigFile()
+{
+    std::ofstream configFile(AGENT_CONFIG_PATH);
```
jr0me, line 20: Will this overwrite the file if it already exists?

src/cmake/CommonSettings.cmake
```diff
@@ -3,6 +3,35 @@ function(set_common_settings)
     set(CMAKE_CXX_STANDARD_REQUIRED ON PARENT_SCOPE)
     set(CMAKE_EXPORT_COMPILE_COMMANDS ON PARENT_SCOPE)
 
+    if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
jr0me, line 6: This will be called for every target that uses this function, but it doesn't seem to be needed... maybe we should create a separate file with a new function to do the installation related setup, and have this function only be called from the root cmake file.

src/agent/multitype_queue/include/multitype_queue.hpp
TomasTurina: Same for the command store DB.

jr0me (commented):
I commented on a few minor things. It has become clear that we need to mock the database operations to improve all these tests. But overall LGTM.

src/agent/multitype_queue/include/multitype_queue.hpp
vikman90: @TomasTurina We don't need that as we always need to define one at code level. The configuration parser will provide a default value, but neither the Queue nor the Command Store need to apply a default value.

cborla:
The PR looks good, let's keep in mind Tomás' [comment](https://github.com/wazuh/wazuh-agent/pull/325#discussion_r1856751572).

PR #344, opened 2024-11-22
Inventory Network Tables Unification

### LINKED ISSUE #284: Refactor network interfaces and protocols inventory tables
**Parent Issue:** #292

The current network inventory separates interfaces and protocols, since each network interface may have zero or many IP addresses.

```mermaid
erDiagram
    INTERFACE ||--o{ PROTOCOL : has
    PROTOCOL ||--|{ ADDRESS : has
```

Now, since we're synchronizing this data with Indexer (which is not an SQL database), we need to merge both tables into one, repeating the data originally stored in the interfaces table.

Take into account that, since the relationship is zero-or-many to one, we must ensure that there is at less one entry for each network interface, even if it has no IP addresses.

## Depends on

- Agree indexer definition with @wazuh/devel-indexer-leaders.
   - https://github.com/wazuh/wazuh-indexer/issues/270#issuecomment-2461850799

LucioDonda:
|Related issue|
|---|
|Closes #284 |

## Description

Before remapping to the ECS compliant field we are looking for a way of unifying the network data.
This implies erging network address iface and protocols into a single table called network.


src/modules/inventory/src/inventoryImp.cpp
```diff
@@ -618,87 +595,63 @@ nlohmann::json Inventory::GetNetworkData()
...
+                        networkTableData["network_checksum"]   = GetItemChecksum(networkTableData);
+                        networkTableData["network_item_id"]    = GetItemId(networkTableData, NETWORK_ITEM_ID_FIELDS);
+
+                        ret[NETWORKS_TABLE].push_back(networkTableData);
                     }
                 }
             }
-
-            ret[NET_IFACE_TABLE] = std::move(ifaceTableDataList);
-            ret[NET_PROTOCOL_TABLE] = std::move(protoTableDataList);
```
cborla, line 700: Is it possible to have an unaddressed interface?

LucioDonda: AFAIK there's no chance of getting an interface without an address, in those cases the data provider is throwing an exception. Example on sysInfoLinux.cpp:

```cpp
        nlohmann::json ifaddr {};

        for (auto addr : interface.second)
        {
            const auto networkInterfacePtr { FactoryNetworkFamilyCreator<OSPlatformType::LINUX>::create(std::make_shared<NetworkLinuxInterface>(addr)) };

            if (networkInterfacePtr)
            {
                networkInterfacePtr->buildNetworkData(ifaddr);
            }
        }

        networks["iface"].push_back(ifaddr);
```

```cpp
void LinuxNetworkImpl<AF_INET>::buildNetworkData(nlohmann::json& network)
{
    // Get IPv4 address
    const auto address { m_interfaceAddress->address() };

    if (!address.empty())
    {
        nlohmann::json ipv4JS { };
        ipv4JS["address"] = address;
        ipv4JS["netmask"] = m_interfaceAddress->netmask();
        ipv4JS["broadcast"] = m_interfaceAddress->broadcast();
        ipv4JS["metric"] = m_interfaceAddress->metrics();
        ipv4JS["dhcp"]   = m_interfaceAddress->dhcp();

        network["IPv4"].push_back(ipv4JS);
    }
    else
    {
        throw std::runtime_error { "Invalid IpV4 address." };
    }
}
```

cborla (approved):
LGTM

PR #350, opened 2024-11-25
Race condition check for if file was deleted

JakeHuneau:
It is possible a file is deleted between the initial check and the `lstat` call on the file which will incorrectly flag it as a potential root kit.

One common example that I see is for a file like `/etc/passwd.lock` where I believe the lock file is deleted between the call to `read_sys_file` and the `lstat` call. Checking for the `errorno` after `lstat` can indicate if the file doesn't exist and then double checking that the file really doesn't exist with `access` can remediate this race condition.

## Description

Checks if a file was deleted during a race condition between a file being scanned and `lstat` being checked on the file.

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X
- [ ] Source installation
- [ ] Package installation
- [ ] Source upgrade
- [ ] Package upgrade
- [ ] Review logs syntax and correct language
- [ ] QA templates contemplate the added capabilities

- Memory tests for Linux
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Valgrind (memcheck and descriptor leaks check)
  - [ ] Dr. Memory
  - [ ] AddressSanitizer
- Memory tests for Windows
  - [ ] Scan-build report
  - [ ] Coverity
  - [ ] Dr. Memory
- Memory tests for macOS
  - [ ] Scan-build report
  - [ ] Leaks
  - [ ] AddressSanitizer

### DIFF
--- a/src/modules/rootcheck/src/check_rc_sys.c
+++ b/src/modules/rootcheck/src/check_rc_sys.c
@@ -10,6 +10,8 @@
 
 #include "shared.h"
 #include "rootcheck.h"
+#include <errno.h>
+#include <unistd.h>
 
 /* Prototypes */
 static int read_sys_file(const char *file_name, int do_read);
@@ -35,6 +37,14 @@ static int read_sys_file(const char *file_name, int do_read)
     os_check_ads(file_name);
 #endif
     if (lstat(file_name, &statbuf) < 0) {
+	// Check if the specified error is that the file no longer exists
+        if (errno == ENOENT) {
+            // Double-check file existence to handle potential race condition
+            if (access(file_name, F_OK) != 0) {
+                // File definitely does not exist, likely removed between checks
+                return 0;
+            }
+        }
 #ifndef WIN32
         const char op_msg_fmt[] = "Anomaly detected in file '%*s'. Hidden from stats, but showing up on readdir. Possible kernel level rootkit.";
         char op_msg[OS_SIZE_1024 + 1];


JakeHuneau:
I will run the tests later today once I'm able.

JakeHuneau:
I tried building I tried to build this on Mac and Ubuntu following the [build instructions](https://github.com/wazuh/wazuh-agent/blob/master/BUILD.md) and get the following error:

> CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!

I get the same error on Ubuntu after following the instructions. I've included the full log. Does anyone have guidance on this?

[vcpkg-manifest-install.log](https://github.com/user-attachments/files/17909869/vcpkg-manifest-install.log)

jr0me:
Hi @JakeHuneau 

The error you're getting is

```
error: Missing GNOME-libxml2-v2.11.7.tar.gz and downloads are blocked by x-block-origin.
error: https://gitlab.gnome.org//GNOME/libxml2/-/archive/v2.11.7/libxml2-v2.11.7.tar.gz: curl failed to download with exit code 56
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

curl: (56) The requested URL returned error: 503
```

And hints at a problem trying to download one of the dependencies, I would suggest to verify your connection to that repository and try again.

Furthermore, are you sure you want to use this version of the Wazuh Agent? This repository is for the version 5.0 which is currently under development, if you are need one of the released versions you need to check out another repository https://github.com/wazuh/wazuh

jr0me:
@JakeHuneau you can target the branch for the version in which you found the bug, or master, whoever reviews your PR will eventually decide where it needs to be merge. Any changes made to prior versions will be ported to the newer versions if it makes sense. As for timelines, there are no specific dates yet, but we are aiming at 2025.

Issue #345, opened 2024-11-25
Define new location for wazuh-agent.lock file


sdvendramini:
**Parent Issue:** https://github.com/wazuh/wazuh-agent/issues/241

The Wazuh agent currently creates the wazuh-agent.lock file in the execution directory under var/run.

Further research will be needed to determine the proper location for this file, ensuring compatibility across all platforms.

sdvendramini:
### 25/11/2024
I have started researching to define the new location and performed the first tests.

PR #352, opened 2024-11-26
Define new location for wazuh-agent.lock file

sdvendramini:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/345|

## Description
The path used to create the `wazuh-agent.lock` file has been parameterized and is set by default to match the `DATA_INSTALL_DIR` path. This value can be modified at any time in `src/cmake/config.cmake`.


src/agent/src/unix_daemon.cpp
```diff
@@ -91,51 +89,14 @@ namespace unix_daemon
...
     {
-        return {};
+        return {LockFileHandler(std::move(lockFilePath))};
     }
 
-    std::string GetDaemonStatus()
+    std::string GetDaemonStatus(std::string lockFilePath)
     {
-        LockFileHandler lockFileHandler = GenerateLockFile();
+        LockFileHandler lockFileHandler = GenerateLockFile(std::move(lockFilePath));
```
aritosteles, line 99: Shouldn´t LockFileHandler remove the LockFile on destruction?

sdvendramini: I did not add the file deletion in the destructor of the function because when a new lock is created or its status is checked, instances of this object are also created, causing issues when attempting to delete the file. Instead, I directly delete the file when the agent stops running properly in [here](https://github.com/wazuh/wazuh-agent/blob/4923ec1840be0c2c82901168ffac8f2a306ac6f3/src/agent/src/process_options_unix.cpp#L30.) . What do you think?

src/agent/CMakeLists.txt
```diff
@@ -73,6 +73,7 @@ target_link_libraries(Agent
     nlohmann_json::nlohmann_json
     fmt::fmt
     Logger
+    Config
```
TomasTurina, line 76: Why wasn't this needed before?

src/agent/src/unix_daemon.cpp
jr0me: So what happens if the agent crashes and the lock file is not removed? Sorry I know the question may not be 100% related to this PR, but I lost track of this part.

src/agent/CMakeLists.txt
jr0me: Some other target used by the Agent may have linked against Config publicly.

src/agent/src/unix_daemon.cpp
aritosteles: I'm asking because the original idea, before lockfiles (when this was PIDFIleHandler), was to delete the file on shutdown, RAII style. Status checking function was static to avoid creating an object. I don´t know if that still makes sense, but if it does, it's always better. I'm fine either way.

sdvendramini: @jr0me If the agent crashes, the lock on the file disappears, allowing a new agent instance to lock the same file again.

Regarding where the file is deleted, @aritosteles, the issue arises because a new agent instance could open a file descriptor for the file but fail to acquire the lock. It would then close the file, causing the LockFileHandler instance to disappear, resulting in an error when attempting to delete a file it does not own, or potentially deleting it (I should test this particular scenario).

src/agent/src/process_options.hpp
```diff
@@ -57,7 +57,7 @@ void RestartAgent(const std::string& configFile);
 void StartAgent(const std::string& configFile);
 
 /// @brief Displays the current status of the agent.
-void StatusAgent();
+void StatusAgent(const std::string& configFile);
```
TomasTurina, line 60: Add parameter to function description.

src/agent/src/unix_daemon.hpp
```diff
@@ -13,7 +13,7 @@ namespace unix_daemon
     public:
         /// @brief Default constructor
         /// @details Creates a lock file when the instance is created
-        LockFileHandler();
+        LockFileHandler(std::string lockFilePath);
```
TomasTurina, line 16: Add parameter to constructor description.

src/agent/tests/unix_daemon_test.cpp
```diff
@@ -2,26 +2,41 @@
...
+    {
+        m_tempConfigFilePath = "temp_wazuh-agent.yml";
+
+        std::ofstream outFile(m_tempConfigFilePath);
+        outFile << R"(
+            agent:
+                path.run: "."
+        )";
+        outFile.close();
+    }
```
jr0me, line 23: Isn't it possible to only create this file once? The `SetUp` method is called in between every test in the fixture. You could also do it in the constructor of `UnixDaemonTest`.

Who deletes this file?

src/agent/src/process_options.hpp
sdvendramini: Done!

src/agent/src/unix_daemon.hpp
sdvendramini: Done!

src/agent/tests/unix_daemon_test.cpp
sdvendramini: Fixed! I replace the functions SetUp() and TearDown() with SetUpTestSuite() and TearDownTestSuite().

TomasTurina (approved):

aritosteles (approved):

PR #355, opened 2024-11-26
Fix agent reconnection issues and timing logic corrections

### LINKED ISSUE #354: Agent fails to retry connection after server disconnection
## Description

The Wazuh Agent hangs indefinitely when the server becomes unavailable. This issue occurs despite the expectation that the agent should wait for `retry_interval` (30 seconds by default) and attempt to reconnect. 

## Root Cause
Upon investigation, we discovered that when the agent closes a socket, the underlying Boost library throws a `boost::system::system_error` exception by definition. This exception is not handled, resulting in the termination of the coroutine responsible for managing the reconnection attempts.

## Steps to Reproduce

1. Start the agent and connect it to a server.
2. Terminate the server while the agent is connected.
3. Observe that the agent does not attempt to reconnect after the disconnection.

## Expected Behavior
The agent should wait for the configured `retry_interval` and attempt to reconnect to the server.

## Actual Behavior
The agent does not attempt to reconnect and remains idle indefinitely.

## Proposed Solution
Handle and ignore the `boost::system::system_error` exception to prevent the coroutine from terminating. This ensures that the agent can continue its reconnection logic as expected.

vikman90:
|Related issue|
|---|
| Closes #354|

## Description

This PR resolves issues related to the agent's behavior during server disconnections and introduces improvements in timing calculations.

## Key Changes

1. **Avoid socket closure during incomplete handshake:**
   - Modified the logic to ensure the connection socket is not prematurely closed if the TLS handshake has not been completed. This prevents unnecessary errors and ensures a cleaner reconnection process.
2. **Handle exceptions:**
   - Added handling for exceptions that may be thrown by socket and resolver methods. If such an exception occurs, it is now logged as an error instead of causing the coroutine to terminate.
3. **Corrected timing multipliers:**
   - Fixed inaccuracies in time multipliers used during reconnection logic to ensure the agent adheres to the specified `retry_interval`.
4. **Added a configuration option for the concurrency level**
   - On systems there `std::thread::hardware_concurrency` returned small values, the Agent wasn't working as expected due to the resource constraint on the number of threads.

### Motivation

These changes address critical issues where the agent would stop retrying after server disconnection due to unhandled exceptions or incorrect handling of connection states. Additionally, they ensure the timing behavior aligns with user expectations and configuration.

## Testing

- [ ] **Server Disconnection Scenarios:**
   - Tested agent behavior when the server disconnects before and after the handshake is completed.
   - Verified the agent retries reconnection according to the configured `retry_interval`.
- [ ] **Exception Handling:**
   - Simulated socket closure errors and confirmed they are logged without terminating the coroutine.
- [ ]  **Timing Validation:**
   - Ensured the agent's reconnection attempts are delayed by exactly the configured `retry_interval`.


TomasTurina (commented):
Missing some UT.

src/agent/communicator/include/ihttp_socket.hpp
```diff
@@ -17,7 +17,8 @@ namespace http_client
 
         /// @brief Connects the socket to the given endpoints
         /// @param endpoints The endpoints to connect to
-        virtual void Connect(const boost::asio::ip::tcp::resolver::results_type& endpoints) = 0;
+        virtual void Connect(const boost::asio::ip::tcp::resolver::results_type& endpoints,
```
TomasTurina, line 20: Add new parameter to functions descriptions.

src/agent/communicator/src/http_client.cpp
TomasTurina, line 1: Validate `socket` and `results` in `Co_PerformHttpRequest`.

src/agent/communicator/src/http_client.cpp
```diff
@@ -125,7 +124,7 @@ namespace http_client
```
TomasTurina, line 124: In line https://github.com/wazuh/wazuh-agent/pull/355/files#diff-301a5da8240724f147370f285f68ff252a29505ea09942939dbec96f24034596R120, make `req` `const`.

src/agent/communicator/src/http_client.cpp
```diff
@@ -101,8 +101,7 @@ namespace http_client
                         reqParams.Endpoint,
```
TomasTurina, line 101: Maybe the variable `code` from https://github.com/wazuh/wazuh-agent/pull/355/files#diff-301a5da8240724f147370f285f68ff252a29505ea09942939dbec96f24034596R95 can be reused as you did in `PerformHttpRequest`.

TomasTurina: What is the difference between `boost::system::error_code` and `boost::beast::error_code`?

src/agent/communicator/src/http_client.cpp
TomasTurina: The same in `PerformHttpRequestDownload`.

src/agent/communicator/src/http_client.cpp
```diff
@@ -282,10 +311,23 @@ namespace http_client
...
-            socket->Write(req);
+
+            socket->Write(req, ec);
+
+            if (ec)
+            {
+                throw std::runtime_error("Error writing request: " + ec.message());
+            }
+
             socket->ReadToFile(res_parser, dstFilePath);
```
TomasTurina, line 331: No error code check here?

TomasTurina: The code of this function is pretty similar than `PerformHttpRequest`. Maybe we can create a private function for it?

src/agent/communicator/src/https_socket.hpp
```diff
@@ -37,18 +58,40 @@ namespace http_client
...
+                                                    boost::asio::redirect_error(boost::asio::use_awaitable, code));
 
-            co_await m_ssl_socket.async_handshake(boost::asio::ssl::stream_base::client,
-                                                  boost::asio::redirect_error(boost::asio::use_awaitable, code));
+                if (code)
+                {
+                    LogDebug("boost::asio::async_connect returned error code: {} {}", code.value(), code.message());
+                }
+
+                co_await m_ssl_socket.async_handshake(boost::asio::ssl::stream_base::client,
```
TomasTurina, line 72: Check for `code`?

src/agent/communicator/src/http_client.cpp
```diff
@@ -154,7 +153,7 @@ namespace http_client
                 {
                     onUnauthorized();
                 }
-                timerSleep = connectionRetry * A_SECOND_IN_MILLIS;
+                timerSleep = connectionRetry;
             }
 
             LogDebug("Response code: {}.", res.result_int());
```
TomasTurina, line 159: Add also endpoint name.

src/agent/communicator/src/http_client.cpp
```diff
@@ -178,19 +177,49 @@ namespace http_client
...
+            }
+
+            socket->Read(res, ec);
+
+            if (ec)
+            {
+                throw std::runtime_error("Error reading response: " + ec.message());
+            }
 
             LogDebug("Response code: {}.", res.result_int());
```
TomasTurina, line 217: Add also endpoint name.

src/agent/communicator/src/http_client.cpp
```diff
@@ -282,10 +311,23 @@ namespace http_client
...
+            socket->Write(req, ec);
+
+            if (ec)
+            {
+                throw std::runtime_error("Error writing request: " + ec.message());
+            }
+
             socket->ReadToFile(res_parser, dstFilePath);
 
             LogDebug("Response code: {}.", res_parser.get().result_int());
```
TomasTurina, line 333: Add also endpoint name.

src/agent/communicator/src/http_client.cpp
```diff
@@ -86,25 +106,40 @@ namespace http_client
 
         do
         {
-            std::time_t timerSleep = A_SECOND_IN_MILLIS;
+            if (!token || token->empty())
+            {
+                co_await WaitForTimer(timer, connectionRetry);
```
TomasTurina, line 111: I think we should sleep a short period of time here, maybe 1s is enough.

jr0me (approved):
LGTM :100:

TomasTurina (approved):
GJ!

sdvendramini (approved):

Issue #154, opened 2024-09-13
Mark in-progress commands as not executed after restarting an agent


TomasTurina:
## Description

As part of the development of the new agent, a [command handler](https://github.com/wazuh/wazuh-agent/issues/83) was introduced to handle the execution of commands received from the server.

Since an agent can stop or restart its execution at any time, this can happen while there are commands in progress that have not yet reported their result. Since the manager needs to know the result of each command, the command handler must check if there are any commands in progress at startup and report their result to the manager. For these cases, the commands must be marked as "agent stopped during execution" and this will be the reported result.

## Considerations

- The command handler should check for in-progress commands upon startup.
- This will be a one-time check.
- If in-progress commands are found, their status and result will be updated in the command store to reflect that the agent stopped during execution.
- The result should be added to the queue to inform the manager, which will be waiting for this response.
- If specified, the command should be removed from the command store after the result is queued.

sdvendramini:
### 28/11/2024
I've started working on the issue, analyzing the existing code and considering possible changes needed to achieve the issue's goal.

sdvendramini:
As agreed with @TomasTurina @GGP1 @AlexRuiz7 , the commands will have a new field called `document_id`, which will act as their unique identifier. They will be structured as follows:

```
  "commands": [
    {
      "document_id" : "SuyTlZIBRVsFsWNvvx6f",
      "request_id": "aeyvlZIBRVsFNWNvvj6f",
      "order_id": "auyvlZIBRVsFNWNvvj6f",
      "source": "Users/Services",
      "user": "Management API",
      "target": {
        "id": "0a96a0ab-5bef-415c-bb3c-ea3e294215a0",
        "type": "agent"
      },
      "action": {
        "name": "restart",
        "version": "v5.0.0",
        "args": []
      },
      "timeout": 100,
      "status": "pending"
    }
]
```

Stateful events reporting the command results will look like this:

```
{"module": "command", "id": "SuyTlZIBRVsFsWNvvx6f", "operation": "update"}
{"command":{"result":{"code":4,"message":"Agent stopped during execution"}}}
{"module": "command", "id": "1", "operation": "update"}
{"command":{"result":{"code":0,"message":"Success"}}}
```

PR #391, opened 2024-12-05
Mark in progress commands as not executed after restarting an agent

sdvendramini:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/154|

## Description

This PR introduces the necessary changes to mark commands interrupted by a restart as not executed and to report them correctly to the server.


src/agent/src/agent.cpp
```diff
@@ -42,6 +42,18 @@ Agent::Agent(const std::string& configFilePath, std::unique_ptr<ISignalHandler>
 
     m_taskManager.Start(
         m_configurationParser.GetConfig<size_t>("agent", "thread_count").value_or(config::DEFAULT_THREAD_COUNT));
+
+    m_commandHandler.CleanUpInProgressCommands(
```
TomasTurina, line 46: I think I prefer to do this at the beginning of `CommandsProcessingTask`. If so, function can be private.

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -56,8 +59,18 @@ namespace command_handler
                     continue;
                 }
 
-                m_commandStore.StoreCommand(cmd.value());
                 PopCommandFromQueue();
```
TomasTurina, line 62: This should be after the store.

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -72,6 +85,24 @@ namespace command_handler
...
+        void CleanUpInProgressCommands(
+            std::function<void(const std::optional<std::vector<module_command::CommandEntry>>&)> callback)
+        {
+            auto cmds = m_commandStore.GetCommandByStatus(module_command::Status::IN_PROGRESS);
+
+            if (cmds != std::nullopt)
+            {
+                for (auto& cmd : *cmds)
+                {
+                    cmd.ExecutionResult.ErrorCode = module_command::Status::UNKNOWN;
```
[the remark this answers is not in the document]
sdvendramini: Fixed!

src/agent/command_handler/include/command_handler.hpp
sdvendramini: If, when the store operation fails, we mark the command as failed, it would make no difference whether the pop is called before or after. This way, we avoid having to call it twice. Because the command must be removed from the queue, as it could fail due to a duplicate ID. If we don't remove it, the command might end up being processed indefinitely.

src/agent/src/agent.cpp
sdvendramini: Done!

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -76,6 +93,23 @@ namespace command_handler
         void Stop();
 
     private:
+        template<typename T>
+        void CleanUpInProgressCommands(std::function<void(T&)> ReportCommandResult)
```
TomasTurina, line 97: Add function description.

sdvendramini: done!

TomasTurina (approved):

Issue #401, opened 2024-12-09
Expand Success Responses from 200 to 2xx


cborla:
## Description
Currently, our system only considers HTTP status code `200` as a successful response when interacting APIs. However, the engine API documentation specifies that all `2xx` status codes represent successful operations. This includes:

- `201 Created`
- `202 Accepted`
- `204 No Content`

To improve compatibility and functionality, the system should be updated to handle all `2xx` responses as successful.

## Tasks
1. **Review API Client Code**:
   - Identify the sections of the codebase where success is determined based on the HTTP status code.
   - Modify the logic to include all `2xx` status codes.

2. **Update Tests**:
   - Create or update test cases to validate the handling of other `2xx` responses such as `201`, `202`, and `204`.

## Acceptance Criteria
- All `2xx` status codes are treated as successful responses.
- Existing functionality with `200` responses remains unchanged.

PR #402, opened 2024-12-09
Expand success responses from 200 to 2xx

jr0me:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/401|

## Description

This PR treates all 2xx error codes as success.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

### DIFF
--- a/src/agent/communicator/src/communicator.cpp
+++ b/src/agent/communicator/src/communicator.cpp
@@ -236,7 +236,8 @@ namespace communicator
 
         const auto result = m_httpClient->PerformHttpRequestDownload(reqParams, dstFilePath);
 
-        return result.result() == boost::beast::http::status::ok;
+        return result.result() >= boost::beast::http::status::ok &&
+               result.result() < boost::beast::http::status::multiple_choices;
     }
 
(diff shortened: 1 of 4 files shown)


src/agent/communicator/src/communicator.cpp
```diff
@@ -111,7 +111,8 @@ namespace communicator
                 try
                 {
                     const auto result = SendAuthenticationRequest();
-                    if (result != boost::beast::http::status::ok)
+                    if (result < boost::beast::http::status::ok ||
```
TomasTurina, line 114: I think this is not necessary, function `SendAuthenticationRequest` always returns ok.

src/agent/communicator/src/communicator.cpp
TomasTurina, line 1: Check that the changes don't affect `GetGroupConfigurationFromManager`.

jr0me: Good catch, fixed.

TomasTurina (approved):

Issue #394, opened 2024-12-05
Adapt command structure to new definition


TomasTurina:
**Parent Issue**: #241

## Description

After the latest changes introduced in the server side, it was decided that the structure of the messages that the agent receives with commands will be as follows:

```
  "commands": [
    {
      "document_id" : "SuyTlZIBRVsFsWNvvx6f",
      "request_id": "aeyvlZIBRVsFNWNvvj6f",
      "order_id": "auyvlZIBRVsFNWNvvj6f",
      "source": "Users/Services",
      "user": "Management API",
      "target": {
        "id": "0a96a0ab-5bef-415c-bb3c-ea3e294215a0",
        "type": "agent"
      },
      "action": {
        "name": "restart",
        "version": "v5.0.0",
        "args": []
      },
      "timeout": 100,
      "status": "pending"
    }
]
```

Besides, the structure of the response should be as following:

```
{"module": "command", "id": "SuyTlZIBRVsFsWNvvx6f", "operation": "update"}
{"command":{"result":{"code":4,"message":"Agent stopped during execution"}}}
```

This issue consists in adapt the agent code to these new specifications.

Since the new message doesn't specify a module where the command should be executed, we also need to change the logic and make the command handler responsible of distribute the commands based on the name of the command and the arguments. This will require to make changes in the centralized configuration component as well.

It is mandatory to test all these changes with a real server to verify that the command feature works E2E.

sdvendramini:
## 09/12/2024
The necessary modifications to adapt the code to the new command structure have been started.

sdvendramini:
### Tests with python mock

<details>

<summary>Case 1</summary>

- command

```
{
    "commands":
    [
        {
            "document_id": "1193bf02-fad7-7bae-93c7-51c6fbaf62e2",
            "action":
            {
                "name": "set-group",
                "version": "v5.0.0",
                "args":
                [
                    "validYaml",
                    "invalidYaml"
                ]
            },
            "target":
            {
                "type": "agent",
                "id": "agentID"
            },
            "status": "sent"
        }
    ]
}
```

- logs

```
[2024-12-13 09:07:41.822] [wazuh-agent] [info] [INFO] [command_handler_utils.cpp:77] [DispatchCommand] Dispatching command set-group(CentralizedConfiguration)
[...]

PR #408, opened 2024-12-10
Adapt command structure to new definition

sdvendramini:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/394|

## Description
This PR aims to complete the necessary changes to adapt command processing to the newly defined structure received by the agent.

### DIFF
--- a/src/agent/command_handler/include/command_handler.hpp
+++ b/src/agent/command_handler/include/command_handler.hpp
@@ -61,6 +61,19 @@ namespace command_handler
                     continue;
                 }
 
+                if (!CheckCommand(cmd.value()))
+                {
+                    cmd.value().ExecutionResult.ErrorCode = module_command::Status::FAILURE;
+                    cmd.value().ExecutionResult.Message = "Command is not valid";
+                    LogError("Error checking module and args for command: {} {}. Error: {}",
+                             cmd.value().Id,
+                             cmd.value().Command,
+                             cmd.value().ExecutionResult.Message);
+                    ReportCommandResult(cmd.value());
+                    PopCommandFromQueue();
+                    continue;
+                }
+
                 if (!m_commandStore.StoreCommand(cmd.value()))
                 {
                     cmd.value().ExecutionResult.ErrorCode = module_command::Status::FAILURE;
@@ -117,6 +130,18 @@ namespace command_handler
                 }
             }
         }
+
+        /// @brief Check if the command is valid
+        ///
+        /// This function checks if the given command is valid by looking it up in
+        /// the map of valid commands. If the command is valid, it checks if the
+        /// parameters are valid. If the command is not valid, it logs an error and
+        /// returns false. If the command is valid, it sets the module for the
+        /// command and returns true.
+        ///
+        /// @param cmd The command to check
+        /// @return True if the command is valid, false otherwise
+        bool CheckCommand(module_command::CommandEntry& cmd);
 
         /// @brief Indicates whether the command handler is running or not
         std::atomic<bool> m_keepRunning = true;
--- a/src/agent/src/command_handler_utils.cpp
+++ b/src/agent/src/command_handler_utils.cpp
@@ -13,32 +13,32 @@ namespace
     template<typename ExecuteFunction>
     boost::asio::awaitable<void> ExecuteCommandTask(ExecuteFunction executeFunction,
                                                     module_command::CommandEntry commandEntry,
-                                                    std::shared_ptr<module_command::CommandExecutionResult> result,
-                                                    std::shared_ptr<bool> commandCompleted,
-                                                    std::shared_ptr<boost::asio::steady_timer> timer)
+                                                    std::shared_ptr<module_command::CommandExecutionResult> result)
     {
         try
         {
             *result = co_await executeFunction(commandEntry.Command, commandEntry.Parameters);
-            *commandCompleted = true;
-            timer->cancel();
         }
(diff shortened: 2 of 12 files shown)


src/tests/mock-server/config/commands.groovy
```diff
@@ -0,0 +1,50 @@
...
+def numCommands = new Random().nextInt(3)
+
+def commands = []
+if (numCommands > 0) {
+    for (int i = 0; i < numCommands; i++) {
+        def action = actions[new Random().nextInt(actions.size())]
+        def command = [
+            "document_id": generateUUIDv7(),
+            "action": action,
+            "target": ["type": "agent", "id": "agentID"],
```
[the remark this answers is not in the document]
sdvendramini: It's just a trivial value; it doesn't need to be edited since the mock will send all commands to the agent that made the request.

src/agent/src/message_queue_utils.cpp
```diff
@@ -62,37 +62,30 @@ std::optional<module_command::CommandEntry> GetCommandFromQueue(std::shared_ptr<
...
-                    break;
-                default: parameters.push_back(arg); break;
+                parameters.push_back(arg);
             }
         }
     }
 
-    module_command::CommandEntry cmd(id, module, command, parameters, "", module_command::Status::IN_PROGRESS);
-
+    module_command::CommandEntry cmd(id, "", command, parameters, "", module_command::Status::IN_PROGRESS);
```
TomasTurina, line 88: Is the `cmd.Module` still being used? If not, please remove it and all its usages.

TomasTurina: I see you still use it in later stages so I think we can keep it.

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -118,10 +137,54 @@ namespace command_handler
...
+        ///
+        /// This function checks if the given command is valid by looking it up in
+        /// the map of valid commands. If the command is valid, it checks if the
+        /// parameters are valid. If the command is not valid, it logs an error and
+        /// returns false. If the command is valid, it sets the module for the
+        /// command and returns true.
+        ///
+        /// @param cmd The command to check
+        /// @return True if the command is valid, false otherwise
+        bool CheckCommand(module_command::CommandEntry& cmd)
```
TomasTurina, line 150: This function can be in the `cpp` file.

TomasTurina: Same about `CommandInfo` and `m_validCommandsMap`.

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -118,10 +137,54 @@ namespace command_handler
...
+        {
+            auto it = m_validCommandsMap.find(cmd.Command);
+            if (it != m_validCommandsMap.end())
+            {
+                if (it->second.ValidateParams && !cmd.Parameters.empty())
```
TomasTurina, line 155: In this case, shouldn't we check if parameters are not empty?

src/agent/command_handler/include/command_handler.hpp
```diff
@@ -118,10 +137,54 @@ namespace command_handler
...
+                {
+                    for (const auto& param : cmd.Parameters.items())
+                    {
+                        if (!param.value().is_string())
```
TomasTurina, line 159: Let's validate this with @wazuh/devel-indexer and @wazuh/devel-pyserver.

src/tests/mock-server/config/commands.groovy
```diff
@@ -0,0 +1,50 @@
...
+    ["name": "set-group", "version": "v5.0.0", "args": []],
+    ["name": "set-group", "version": "v5.0.0", "args": [""]],
+    ["name": "set-group", "version": "v5.0.0", "args": ["validYaml", 8]],
+    ["name": "set-group", "version": "v5.0.0", "args": ["validYaml"]],
+    ["name": "update-group", "version": "v5.0.0", "args": ["noNeedArgs"]],
+    ["name": "update-group", "version": "v5.0.0"],
+    ["name": "update-group", "version": "v5.0.0", "args": ""]
+]
+
+def numCommands = new Random().nextInt(3)
```
jr0me, line 24: I was testing this branch together with one I'm working on and after setting `numCommands` to a 1000 I got some segmentation faults. Then I checked only in this branch and it also happens. Looks like is not related to the branch though, but just exposed by it. I want to make sure that setting this variable to a high value is valid and that my test is ok, what do you think?

jr0me (approved):
LGTM

src/agent/src/command_handler_utils.cpp
```diff
@@ -13,48 +13,51 @@ namespace
...
     }
 
-    boost::asio::awaitable<void> TimerTask(std::shared_ptr<boost::asio::steady_timer> timer,
-                                           std::shared_ptr<module_command::CommandExecutionResult> result,
-                                           std::shared_ptr<bool> commandCompleted)
+    boost::asio::awaitable<void> TimerTask(std::shared_ptr<module_command::CommandExecutionResult> result)
     {
         try
         {
+            const auto timeout = std::chrono::minutes(60);
```
TomasTurina, line 36: Maybe we can use a `constexpr` for this (60).

sdvendramini: Done!

TomasTurina (approved):

Issue #363, opened 2024-11-28
Make agent modules have their own threads


TomasTurina:
**Parent Issue:** #241

## Description

Currently, modules are designed to run in their own thread. However, they are executed by the task manager together with the coroutines that run the agent's core components. This causes the modules to run as a core component instead of a separate thread to handle its own functions as expected.

To solve this, the proposed solution is to launch each module as a separate thread capable of handling its own tasks and running in a separate location from the main agent components. This solution fits better with the current design of modules and has its benefits, as for example this will allow to easily stop the execution of modules, including their own coroutines.

jr0me:
### Update

Started looking into the issue and its requirements. Will make a first attempt using the TaskManager as data members for the modules.

jr0me:
## Update

Created branch, removed timers from task manager tests, added tests for the thread counter, improved the enqueuing of tasks by enclosing them in try/ctatch blocks. Tested these changes.

jr0me:
## Update

Added a TaskManager to the ModuleManager. Modules threads/coroutines now don't share a context with the Agent, but they still come from an unique pool. If each module needs to have its own context and pool of threads, they'll need a TaskManager as well. Will test this and assess.

jr0me:
## Update

Tested the ModuleManager having its own TaskManager and sharing it with the modules. So far no problems where found.

jr0me:
## Update

Ran more tests, code is ready for code review.

jr0me:
## Update

Dealing with some crashes in tests, caused by the changes in the TaskManager and ModuleManager.

jr0me:
## Update

I'm still debugging the crashes and deadlocks. These tests take time since occurrences are random. I was able to simplify the tests so they don't fail anymore but I've found a segmentation fault when executing commands (by including https://github.com/wazuh/wazuh-agent/pull/408 ).

I've been trying to find the cause with valgrind, but it seems that the segmentation fault doesn't occur as often when run under this tool.

jr0me:
## Update

I've located the segmentation fault, which came from the command dispatch coroutines, specifically the timeout cancellation mechanism. Fixed and merged.

PR #409, opened 2024-12-10
Separate Agent thread context from ModuleManager

jr0me:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/363|

## Description

This PR gives the ModuleManager its own TaskManager so it can EnqueueTasks without interfering with the core Agent tasks. Modules will now use this other task manager for their own tasks too.

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X


src/agent/task_manager/include/itask_manager.hpp
```diff
@@ -24,9 +25,11 @@ class ITaskManager
 
     /// @brief Enqueues a task to be executed
     /// @param task The task to enqueue
-    virtual void EnqueueTask(std::function<void()> task) = 0;
+    /// @param taskID The ID of the task
+    virtual void EnqueueTask(std::function<void()> task, const std::string& taskID) = 0;
```
TomasTurina, line 29: Why did you choose a string for this?

src/agent/task_manager/src/task_manager.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+    auto taskWithThreadCounter = [this, taskID, task = std::move(task)]() mutable
+    {
+        try
+        {
+            task();
+        }
+        catch (const std::exception& e)
+        {
+            LogError("{} task threw an exception: {}", taskID.empty() ? "Anonymous" : taskID, e.what());
```
TomasTurina, line 57:
```suggestion
            LogError("{} task threw an exception: {}. Exiting...", taskID.empty() ? "Anonymous" : taskID, e.what());
```
Or maybe relaunch the task. What do you think?

src/agent/task_manager/include/itask_manager.hpp
jr0me: I thought it was the easiest to set up from a user point of view, but any suggestion is welcome. What I wanted it to be able to know which task threw an exception, which is not possible to do without an identifier.

src/modules/src/moduleManager.cpp
```diff
@@ -33,9 +31,11 @@ std::shared_ptr<ModuleWrapper> ModuleManager::GetModule(const std::string & name
 }
 
 void ModuleManager::Start() {
-    for (const auto &[_, module] : m_modules) {
-        m_createTask([module]() { module->Start(); });
+    for (const auto &[_, module] : m_modules)
+    {
+        m_taskManager.EnqueueTask([module]() { module->Start(); }, module->Name());
     }
+    m_taskManager.Start(m_modules.size() * 2);
```
TomasTurina, line 38:
```suggestion
    m_taskManager.Start(m_modules.size());
```
We only need one thread per module.

src/agent/task_manager/include/itask_manager.hpp
TomasTurina: I agree, let's keep this.

src/agent/task_manager/src/task_manager.cpp
jr0me: I wouldn't try to relaunch the task because we don't know what the task is, besides how many retries should we attempt for a failing task? I think it raises too many questions to do it without research.

src/modules/src/moduleManager.cpp
```diff
@@ -33,9 +31,11 @@ std::shared_ptr<ModuleWrapper> ModuleManager::GetModule(const std::string & name
 }
 
 void ModuleManager::Start() {
-    for (const auto &[_, module] : m_modules) {
-        m_createTask([module]() { module->Start(); });
+    for (const auto &[_, module] : m_modules)
+    {
+        m_taskManager.EnqueueTask([module]() { module->Start(); }, module->Name());
     }
+    m_taskManager.Start(m_modules.size());
```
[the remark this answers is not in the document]
jr0me: Yes, both work! We can make them match if you prefer.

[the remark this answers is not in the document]
TomasTurina: Please! 😄

TomasTurina (approved):

Issue #157, opened 2024-09-13
Improve error handling in network functions


TomasTurina:
## Description

As part of the development of the communicator component of the new agent, we chose to use boost asio/beast as our networking library, which also works with C++20 coroutines.

The first implementation is missing to check for many network errors that can occur during communication. Currently, any failure is not reported correctly and no recovery attempt is made, so we need to improve error handling.

Tasks to be done:
- [x] Ensure that all functions that use the Boost networking library are within the HTTPClient class.
- [x] Identify functions within the HTTPClient class that do not have error handling.
- [x] Apply changes to improve error handling.
- [x] Ensure that components that use the HTTPClient class (i.e. Communicator) are properly handling network errors/exceptions.

## Errors reported during testing

When performing manual tests for other developments we have already encountered some cases that we have to solve:

### Handle jwt exceptions
In the following [line ](https://github.com/wazuh/wazuh-agent/blob/master/src/agent/communicator/src/communicator.cpp#L59)

`if (const auto decoded = jwt::decode<jwt::traits::nlohmann_json>(*m_token); decoded.has_payload_claim("exp"))`

a case occurred where a token was returned with an invalid jwt format and the co-routine that handled this task crashed, causing no further attempt to obtain the token and no errors were logged.

### Chrash when getting reference from null object
In the following [line](https://github.com/wazuh/wazuh-agent/blob/master/src/agent/communicator/src/http_client.cpp#L215)

`return nlohmann::json::parse(boost::beast::buffers_to_string(res.body().data())).at("token").get_ref<const std::string&>();`

if the json object has a different format than expected causing “token” to not be where you are looking for it, it causes the crash when trying to do the get_ref() since the at() above would fail.

Example of response that produces the crash:
```
{
"no-token": "value"
}
```

### End of co-routines in case of write/read failure
In [Co_PerformHttpRequest ](https://github.com/wazuh/wazuh-agent/blob/master/src/agent/communicator/src/http_client.cpp#L70)when a failure occurs during [AsyncWrite](https://github.com/wazuh/wazuh-agent/blob/master/src/agent/communicator/src/http_client.cpp#L125)/[AsyncRead ](https://github.com/wazuh/wazuh-agent/blob/master/src/agent/communicator/src/http_client.cpp#L135)the co_routine is being terminated completely, the correct operation of this should be: 

- close the socket
- do a wait
- try again later

but do not stop the co-routine.

### Timeouts in operations with sockets
[...]

PR #474, opened 2025-01-03
Adds timeouts to all socket operations that can block the process/co_routine indefinitely. 

aritosteles:
|Related issue|
|---|
|[157](https://github.com/wazuh/wazuh-agent/issues/157)|

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

### DIFF
--- a/src/agent/communicator/include/ihttp_socket.hpp
+++ b/src/agent/communicator/include/ihttp_socket.hpp
@@ -8,6 +8,9 @@
 
 namespace http_client
 {
+    /// @brief The socket timeout in seconds
+    constexpr int SOCKET_TIMEOUT_SECS = 60;
+
     /// @brief Interface for HTTP sockets
     class IHttpSocket
     {
--- a/src/agent/communicator/src/http_client.cpp
+++ b/src/agent/communicator/src/http_client.cpp
@@ -242,7 +242,12 @@ namespace http_client
         return PerformHttpRequestInternal(params,
                                           [](std::unique_ptr<IHttpSocket>& socket,
                                              boost::beast::http::response<boost::beast::http::dynamic_body>& res,
-                                             boost::system::error_code& ec) { socket->Read(res, ec); });
+                                             boost::system::error_code& ec,
+                                             boost::asio::io_context& ioContext)
+                                          {
+                                              socket->Read(res, ec);
+                                              ioContext.run();
+                                          });
     }
 
     std::optional<std::string> HttpClient::AuthenticateWithUuidAndKey(const std::string& serverUrl,
@@ -354,7 +359,8 @@ namespace http_client
         const HttpRequestParams& params,
         const std::function<void(std::unique_ptr<IHttpSocket>&,
                                  boost::beast::http::response<boost::beast::http::dynamic_body>&,
-                                 boost::system::error_code&)>& responseHandler)
+                                 boost::system::error_code&,
+                                 boost::asio::io_context&)>& responseHandler)
     {
         boost::beast::http::response<boost::beast::http::dynamic_body> res;
 
@@ -385,6 +391,7 @@ namespace http_client
             boost::system::error_code ec;
 
             socket->Connect(results, ec);
+            io_context.run();
 
             if (ec)
             {
@@ -394,13 +401,14 @@ namespace http_client
             const auto req = CreateHttpRequest(params);
 
             socket->Write(req, ec);
+            io_context.run();
 
             if (ec)
             {
                 throw std::runtime_error("Error writing request: " + ec.message());
             }
 
-            responseHandler(socket, res, ec);
+            responseHandler(socket, res, ec, io_context);
 
             if (ec)
             {
--- a/src/agent/communicator/src/http_client.cpp
+++ b/src/agent/communicator/src/http_client.cpp
@@ -239,15 +239,73 @@ namespace http_client
     boost::beast::http::response<boost::beast::http::dynamic_body>
     HttpClient::PerformHttpRequest(const HttpRequestParams& params)
     {
-        return PerformHttpRequestInternal(params,
-                                          [](std::unique_ptr<IHttpSocket>& socket,
-                                             boost::beast::http::response<boost::beast::http::dynamic_body>& res,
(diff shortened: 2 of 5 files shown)


src/agent/communicator/src/http_socket.hpp
```diff
@@ -151,6 +159,6 @@ namespace http_client
 
     private:
         /// @brief The socket to use for the HTTP connection
-        boost::asio::ip::tcp::socket m_socket;
+        boost::beast::tcp_stream m_beast_socket;
```
TomasTurina, line 162:
```suggestion
        boost::beast::tcp_stream m_socket;
```
Keep previous name.

src/agent/communicator/src/http_socket.hpp
```diff
@@ -40,7 +41,9 @@ namespace http_client
         {
             try
             {
-                boost::asio::connect(m_socket, endpoints, ec);
+                m_beast_socket.expires_after(std::chrono::seconds(http_client::SOCKET_TIMEOUT_SECS));
```
TomasTurina, line 44: Does this function need to be called every time in every function? Or can we set the socket timeout just once as a property?

src/agent/communicator/src/https_socket.hpp
```diff
@@ -69,7 +70,9 @@ namespace http_client
         {
             try
             {
-                boost::asio::connect(m_ssl_socket.next_layer(), endpoints.begin(), endpoints.end(), ec);
+                m_ssl_socket.next_layer().expires_after(std::chrono::seconds(http_client::SOCKET_TIMEOUT_SECS));
```
TomasTurina, line 73: Same question.

src/agent/communicator/include/ihttp_socket.hpp
```diff
@@ -8,6 +8,9 @@
 
 namespace http_client
 {
+    /// @brief The socket timeout in seconds
+    constexpr int SOCKET_TIMEOUT_SECS = 2;
```
TomasTurina, line 12:
```suggestion
    constexpr int SOCKET_TIMEOUT_SECS = 10;
```
I think 2 seconds is a small number, what do you think?

src/agent/communicator/src/http_client.cpp
```diff
@@ -385,6 +392,8 @@ namespace http_client
             boost::system::error_code ec;
 
             socket->Connect(results, ec);
+            size_t ret = io_context.run();
```
TomasTurina, line 395: Why is this necessary?

src/agent/communicator/src/http_client.cpp
```diff
@@ -394,13 +403,15 @@ namespace http_client
             const auto req = CreateHttpRequest(params);
 
             socket->Write(req, ec);
+            ret = io_context.run();
+            LogDebug("io_context run returned {}", ret);
```
TomasTurina, line 407: Remove log.

src/agent/communicator/src/http_client.cpp
```diff
@@ -144,6 +144,7 @@ namespace http_client
 
             boost::system::error_code ec;
 
+            LogDebug("Attempting Async connection: {}", reqParams.Endpoint);
```
TomasTurina, line 147: Remove log.

src/agent/communicator/src/http_socket.hpp
aritosteles: No, it has to be called every time. Otherwise the (internal) timer keeps running since the first call.

src/agent/communicator/src/http_client.cpp
aritosteles: io_context needs to run all outstanding asynchronous operations so that either the internal timer expires or the operation is completed. The result is the operations that were ran and it's logged just for debugging purposes.

src/agent/communicator/src/http_client.cpp
```diff
@@ -385,6 +392,8 @@ namespace http_client
             boost::system::error_code ec;
 
             socket->Connect(results, ec);
+            size_t ret = io_context.run();
+            LogDebug("io_context run returned {}", ret);
```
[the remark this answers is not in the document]
aritosteles: Yes, just for debugging purposes.

src/agent/communicator/src/http_client.cpp
aritosteles: Same

src/agent/communicator/src/http_client.cpp
aritosteles: Same

src/agent/communicator/src/https_socket.hpp
aritosteles: Same here

src/agent/communicator/src/http_client.cpp
```diff
@@ -394,13 +401,14 @@ namespace http_client
...
             socket->Write(req, ec);
+            io_context.run();
 
             if (ec)
             {
                 throw std::runtime_error("Error writing request: " + ec.message());
             }
 
-            responseHandler(socket, res, ec);
+            responseHandler(socket, res, ec, io_context);
```
jr0me, line 411: Is it necessary to pass the io_context as an argument here? Can't we run io_context.run() after calling the response handler like with the calls to socket connect and socket write?

src/agent/communicator/src/http_client.cpp
```diff
@@ -242,7 +242,12 @@ namespace http_client
         return PerformHttpRequestInternal(params,
```
jr0me, line 242: It looks like when we removed `PerformHttpRequestDownload` in https://github.com/wazuh/wazuh-agent/pull/472 we lost the need to have a `PerformHttpRequestInternal` method. Maybe we should erase it altogether.

TomasTurina: I agree.

src/agent/communicator/src/http_client.cpp
aritosteles: Function PerformHttpRequestInternal() has been removed.

TomasTurina (approved):

Issue #468, opened 2024-12-30
Investigate and adapt the server certificate validation using the API provided by the OS on macOS.


Nicogp:
|Related issue|
|---|
| #389 |

## Description

As a complement to issue #389 it is necessary to investigate the use of CA certificates in macOS and if necessary, adapt the code so that the validation of server certificates is performed through the API provided by the OS.

Certificate validation in macOS should follow the same logic as in #389:

- **`full` (default)**:
  - Validates that the server certificate is signed by a trusted CA.
  - Ensures the server hostname matches the certificate's SAN or CN.
- **`certificate`**:
  - Validates that the server certificate is signed by a trusted CA.
  - Does not validate the server hostname.
- **`none`**:
  - Disables all certificate validation.
  - No checks are performed on the certificate's CA signature or the server hostname.
    **Note:** This mode disables critical SSL/TLS security features and is not recommended for production environments.

Nicogp:
## Work update

### 2024/12/30

- I started with the issue
- I set up an environment with macOS sonoma
- I tested the current changes, and as expected it is failing validation since, like Windows, openssl does not access the OS CA store by default.
![Image](https://github.com/user-attachments/assets/c9d3a9e1-88a3-41bd-bdfb-8ce7fb93d23e)

Nicogp:
## Tests

### Setting up the environment

For testing purposes, I first created a `root CA` and its certificate (MyCA.crt) and then I created two certificates:
`serverIP.crt`: created with the valid IP(CN) of the server.
`serverBadIP.crt`: created with a different IP(CN) than the server to force the host validation failure.

A third certificate is used
`server.crt`: signed with a CA other than MyCA to force the verification failure.


**Adding the MyCA.crt certificate to the agent environment**

<details><summary>MacOS</summary>

- Copy the certificate into a folder
- Import the certificate to macOS Keychain Access: 
  - sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain MyCA.crt

</details>


### Test1: server configured with `server.crt` (certificate signed by a CA unknown to the agents)

Expected result: the agent should be able to connect only when `verification_mode==none`

<details><summary>/etc/wazuh-server/wazuh-server.yml</summary>

```yaml
server:
  nodes:
   - master
  node:
    name: manager_01
    type: master
    ssl:
      key:  "/etc/wazuh-server/certs/server.key"
      cert: "/etc/wazuh-server/certs/server.crt"
[...]

PR #476, opened 2025-01-06
Add server certificate validation to MacOS wazuh agent

Nicogp:
|Related issue|
|---|
|#468|

## Description

This PR adds server certificate validation in the MacOS agent.

For server connections when the agent is already registered, a new configuration option is added to control which mode will be used:
```
agent:
  verification_mode: full
```

Possible Values:

- full (default):
  - Validates that the server certificate is signed by a trusted CA.
  - Ensures the server hostname matches the certificate's SAN or CN.
- certificate:
  - Validates that the server certificate is signed by a trusted CA.
  - Does not validate the server hostname.
- none:
  - Disables all certificate validation.
  - No checks are performed on the certificate's CA signature or the server hostname.
  - Note: This mode disables critical SSL/TLS security features and is not recommended for production environments.

To perform the registration of the agent, a new option was added to the CLI “--verification-mode”, the values that can take are the same as for the previous case:

`wazuh-agent --register-agent --user user --password pass --url https://serverIP:55000 --verification-mode certificate`

## Tests

  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

### DIFF
--- a/src/agent/communicator/CMakeLists.txt
+++ b/src/agent/communicator/CMakeLists.txt
@@ -17,6 +17,8 @@ add_library(Communicator src/communicator.cpp src/http_client.cpp src/http_reque
 
 if(WIN32)
     set(VERIFY_UTILS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/https_socket_verify_utils_win.cpp")
+elseif(APPLE)
+    set(VERIFY_UTILS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/https_socket_verify_utils_mac.cpp")
 else()
     set(VERIFY_UTILS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/https_socket_verify_utils_lin.cpp")
 endif()
@@ -33,6 +35,8 @@ target_link_libraries(Communicator PUBLIC Config Boost::asio Boost::beast Boost:
 
 if(WIN32)
     target_link_libraries(Communicator PRIVATE Crypt32)
+elseif(APPLE)
+    target_link_libraries(Communicator PRIVATE "-framework Security" "-framework CoreFoundation")
 endif()
 
 include(../../cmake/ConfigureTarget.cmake)
--- a/src/agent/communicator/CMakeLists.txt
+++ b/src/agent/communicator/CMakeLists.txt
@@ -13,8 +13,6 @@ find_package(Boost REQUIRED COMPONENTS asio beast system url)
 find_package(nlohmann_json CONFIG REQUIRED)
 find_path(JWT_CPP_INCLUDE_DIRS "jwt-cpp/base.h")
 
-add_library(Communicator src/communicator.cpp src/http_client.cpp src/http_request_params.cpp)
-
 if(WIN32)
     set(VERIFY_UTILS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/https_socket_verify_utils_win.cpp")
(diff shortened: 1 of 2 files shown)


src/agent/communicator/src/https_socket_verify_utils_mac.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+        unsigned char* certData = nullptr;
+        int certLen = i2d_X509(cert, &certData);
+        if (certLen <= 0)
+        {
+            LogError("Failed to encode certificate to DER.");
+            return false;
+        }
+
+        cfCertData = CFDataCreate(kCFAllocatorDefault, certData, certLen);
```
jr0me, line 40: We could define this here in place instead of on line 30.

src/agent/communicator/src/https_socket_verify_utils_mac.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (!cert)
+        {
+            LogError("The server certificate could not be obtained.");
+            return false;
+        }
+
+        CFDataRef cfCertData = nullptr;
+
+        unsigned char* certData = nullptr;
+        int certLen = i2d_X509(cert, &certData);
```
jr0me, line 33: Can this be const?

src/agent/communicator/src/https_socket_verify_utils_mac.cpp
```diff
@@ -0,0 +1,126 @@
...
+        if (status != errSecSuccess || !trust)
+        {
+            LogError("Failed to create SecTrust object.");
+            CFRelease(serverCert);
+            return false;
+        }
+
+        // Evaluate certificate trust using SecTrustEvaluateWithError
+        CFErrorRef error = nullptr;
+        bool trustResult = SecTrustEvaluateWithError(trust, &error);
```
jr0me, line 75: This also looks like it could be const.

src/agent/communicator/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ add_library(Communicator src/communicator.cpp src/http_client.cpp src/http_reque
 
 if(WIN32)
     set(VERIFY_UTILS_FILE "${CMAKE_CURRENT_SOURCE_DIR}/src/https_socket_verify_utils_win.cpp")
+elseif(APPLE)
```
jr0me, line 20: So we don't spread the sources of the target around the file, can we create this "VERIFY_UTILS_FILE" before the call to `add_library` and pass it to it directly after the other files.

jr0me (approved):
LGTM

src/agent/communicator/src/https_socket_verify_utils_mac.cpp
Nicogp: Done!

src/agent/communicator/src/https_socket_verify_utils_mac.cpp
Nicogp: Done!

src/agent/communicator/src/https_socket_verify_utils_mac.cpp
Nicogp: Done!

src/agent/communicator/CMakeLists.txt
Nicogp: Done!

TomasTurina (approved):

Issue #426, opened 2024-12-13
Implement Journald Reader for Logcollector


vikman90:
## Description

Develop a Journald Reader for Logcollector that behaves similarly to the `journalctl` command. The Reader will filter and forward logs based on configurable parameters, ensuring robust integration with the Journald API.

### Configuration Options

The Journald Reader will support the following configuration:

- **`journald`**: Defines a Journald reader with the following options:  
  - **`field`**: Selects a field from Journald entries.  
  - **`regex`**: Regular expression the selected field must match for the log to be transmitted.  
  - **`ignore_if_missing`**: Boolean; defines whether to ignore the log if the specified field does not exist.  

- **`file_wait`**: Defines the log polling interval (e.g., 500ms).  
- **`reload_interval`**: Interval for detecting log rotation (e.g., 1m).

**Example Configuration:**
```yaml
logcollector:
    enabled: true
    reload_interval: 1m
    file_wait: 500ms
    
    journald:
    - field: _SYSTEMD_UNIT
      regex: ^cron.service$
    - field: PRIORITY
      regex: [0-3]
      ignore_if_missing: true
```

### Spike

1. **Review Configuration Options**
   - Ensure that all necessary options are defined and consider additional useful features.
   - Confirm compatibility with Logcollector's configuration structure.
2. **Integrate Journald API**
   - Utilize the client implementation from Wazuh 4.x.
   - Ensure synchronous interaction with the Journald API.
3. **Define Runner Behavior**
   - Design the Journald runner in compliance with the specification outlined in issue #424.
   - Ensure seamless detection of rotation and log state tracking.

### Development

1. **Extend Logcollector's `Setup` Function**
   - Parse and validate the Journald configuration options.
2. **Integrate Journald API Client**
   - Add Journald API as a dependency to facilitate log retrieval and filtering.
3. **Implement Journald Reader**
   - Develop the Reader to process logs based on configured filters.
   - Detect log rotation and handle state persistence for reliable log processing.
4. **Testing**
   - Create unit and integration tests to validate the Reader's behavior under various configurations and scenarios.
5. **Documentation**
   - Provide comprehensive documentation for the Journald Reader, including configuration examples and usage instructions.

## Deliverables

- Fully implemented Journald Reader in Logcollector.
- Unit and integration test coverage for all features.
- Documentation for the Journald Reader.

MarcelKemp:
## Work report

### December 26

- Research current `4.X` code to know all the functionalities. 
[...]

PR #487, opened 2025-01-08
OSLog framework integration

jr0me:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/467|

## Description

This PR introduces a ULS reader for reading logs in macOS.
It uses the OSLogStore class from macOS Foundation SDK, for this, it is implemented using Objective-C++ but wrapped in a C++ class to encapsulate it.

## Tests

- Compilation without warnings in every supported platform
  - [x] Linux
  - [x] Windows
  - [x] MAC OS X

PR #495, opened 2025-01-09
Integration of Journald in Logcollector

MarcelKemp:
|Related issue|
|---|
|#426|

## Description

With this PR we integrate the component to monitor journald using Logcollector.
We have made use of the `sd-journal.h` library, wrapping several functions and adding some extra functionality to add usability.

## Configuration options

To enable integration with **_Journald_**, you need to indicate within `logcollector` that you are going to use it, using `journald:` and it will indicate each condition for monitoring using `-`. This integration has the following configuration options:
- Simple monitoring has the fields:
  - `field:` -> Journald field to be monitored: https://man7.org/linux/man-pages/man7/systemd.journal-fields.7.html
  - `value:` -> Value of the Journald field to be filtered by. You can indicate several values to filter on the same field, using the `|` symbol, which will act as an OR.
    - Ex: For `_SYSTEMD_UNIT`, if you want to filter the _ssh_ or _cron_ service, then the value would be `"ssh.service|cron.service"`. Or for `PRIORITY`, if you want to filter `emerg (0)` messages, then the value would be `"0"`.
  - `exact_match:` -> Boolean that allows the `value` setting to be a substring instead of the exact filtering value (by default = `true`).
  - `ignore_if_missing:` -> Boolean to ignore the filtering condition for logs without the specified field (by default = `false`).
> [!Note] 
> If a substring is given (i.e. `exact_match: false`), no filtering of Journald itself will be applied, being more inefficient and displaying a `warning` indicating this.

- Multiple monitoring (acting as AND): 
```
    - conditions:
        -
        -
```
  - It makes use of `conditions:`, which is given several fields to filter to be applied simultaneously. The filtering settings are the same as for simple monitoring.
    - Ex: If we wanted to filter messages that come from the `cron.service` and also have an `emerg (0)` or `alert (1)` priority, then the configuration would be similar to the following:
```
    - conditions:
        - field: "_SYSTEMD_UNIT"
          value: "cron.service"
        - field: "PRIORITY"
          value: "0|1"
      ignore_if_missing: true
```

#### Example of complete configuration:
```yml
logcollector:
  enabled: true
  localfiles:
    - /var/log/auth.log
  reload_interval: 1m
  file_wait: 500ms

  journald:
    - field: "_SYSTEMD_UNIT"
      value: "cron.service"
      exact_match: true
      ignore_if_missing: true
    - field: "SYSLOG_IDENTIFIER"
      value: "systemd"
      exact_match: false
      ignore_if_missing: true
    - conditions:
        - field: "_COMM"
          value: "cat"
          exact_match: true
        - field: "PRIORITY"
          value: "3|4"
          exact_match: true
      ignore_if_missing: true
```

## Tests

- Compilation in every supported platform
  - [x] Linux
  - [x] Windows
  - [x] Macos


src/modules/logcollector/src/journal_log.hpp
TomasTurina, line 1: Can you move the journald wrapper to a different directory? Check `oslogstore_wrapper` as an example here https://github.com/wazuh/wazuh-agent/pull/487/files.

src/modules/logcollector/src/journal_log.hpp
```diff
@@ -0,0 +1,170 @@
...
+#include <regex>
+#include <optional>
+#include <string_view>
+#include <vector>
+#include <algorithm>
+
+// Forward declaration for sd_journal
+struct sd_journal;
+
+namespace logcollector {
```
TomasTurina, line 16: This namespace may not apply after you relocate this code.

src/modules/logcollector/src/journal_log.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+    /// @brief Gets current epoch time in microseconds
+    static uint64_t GetEpochTime();
+
+    /// @brief Checks and throws on journal operation errors
+    /// @param result Operation result code
+    /// @param operation Operation description for error message
+    void ThrowIfError(int result, const std::string& operation) const;
+
+    /// @brief Applies filter set with OR logic between groups
```
TomasTurina, line 158: Missing parameters in description of these three functions.

src/modules/logcollector/src/logcollector_osx.cpp
```diff
@@ -0,0 +1,12 @@
...
+#include <memory>
+
+namespace logcollector
+{
+
+void Logcollector::AddPlatformSpecificReader(const std::shared_ptr<const configuration::ConfigurationParser>)
+{
+}
+
+} // namespace logcollector
```
TomasTurina, line 12: Missing new line at the end. Check also the other files.

src/modules/logcollector/src/logcollector_unix.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include "journald_reader.hpp"
+
+#include <memory>
+
+namespace logcollector
+{
+
+void Logcollector::AddPlatformSpecificReader(const std::shared_ptr<const configuration::ConfigurationParser> configurationParser)
+{
+    try {
```
TomasTurina, line 12: I'm not sure if this try catch is necessary.

src/modules/logcollector/tests/unit/journal_log_test.cpp
TomasTurina, line 1: Move also tests.

src/modules/logcollector/src/journald_reader.hpp
```diff
@@ -0,0 +1,45 @@
+#pragma once
+
+#include "reader.hpp"
+#include "journal_log.hpp"
```
TomasTurina, line 4:
```suggestion
#include <journal_log.hpp>
```
Once reorganized.

src/modules/logcollector/src/journald_reader.hpp
```diff
@@ -0,0 +1,45 @@
...
+    /// @brief Runs the journal reader
+    /// @return Awaitable for asynchronous operation
+    Awaitable Run() override;
+
+    /// @brief Stops the journal reader
+    void Stop() override;
+
+    /// @brief Gets human-readable description of current filters
+    /// @return String describing current filters
+    std::string GetFilterDescription() const;
```
TomasTurina, line 35: Can be private?

src/modules/logcollector/src/journald_reader.cpp
```diff
@@ -0,0 +1,77 @@
+#include "journald_reader.hpp"
+#include <logger.hpp>
+#include <sstream>
+
+using namespace logcollector;
```
TomasTurina, line 5:
```suggestion
namespace logcollector
{
...
}
```
Why don't you do this?

src/modules/logcollector/src/journald_reader.hpp
MarcelKemp: I had to make it public so I could use it in unit tests.

src/modules/logcollector/tests/unit/logcollector_mock.hpp
```diff
@@ -20,8 +20,18 @@ class LogcollectorMock : public Logcollector {
         Logcollector::SetupFileReader(configurationParser);
     }
 
+    void AddPlatformSpecificReader(std::shared_ptr<const configuration::ConfigurationParser> configurationParser) {
+        Logcollector::AddPlatformSpecificReader(configurationParser);
+    }
+
     MOCK_METHOD(void, AddReader, (std::shared_ptr<IReader> reader), (override));
     MOCK_METHOD(void, EnqueueTask, (Awaitable task), (override));
+    MOCK_METHOD(void, MockWait, (std::chrono::milliseconds));
```
jr0me, line 29: Since you marked `Wait` as virtual, you could override it instead. https://github.com/wazuh/wazuh-agent/pull/495/files#diff-87c624fb50493f22aa7a684c17ff75c5e8d0475232144fa8346c2365827cb7feR64

TomasTurina (approved):

src/modules/logcollector/tests/unit/logcollector_mock.hpp
MarcelKemp: Just below this mock, is the overwritten function that uses it internally.

It is implemented this way because `MockWait` is used to simulate the behaviour of the `Wait` method in tests, allowing to verify that `Wait` is called correctly without introducing real delays.

jr0me: If I understood correctly that's what "EXPECT_CALL" is for. Could it be?

Issue #122, opened 2024-09-10
Unify the usage of the sqlite manager in the new agent


TomasTurina:
## Description

With the development of the new [sqlite_manager](https://github.com/wazuh/wazuh-agent/tree/master/src/agent/sqlite_manager), we have a centralized place to manage database queries.

Currently this component is only being used by `command_store`, but we also need to start using it in the following components:
- [x] `agent_info` related PR: https://github.com/wazuh/wazuh-agent/pull/276
- [ ] `multitype_queue`

This issue consists of starting to use the new database manager in the mentioned components to avoid having duplicate code and centralize access to the databases.

### Considerations

- Rename the `sqlite_manager` component to `persistence`.
- Include a persistence factory.
  - Select the implementation (i.e. SQLite, ...).
- Test the compilation of each component separately.

jr0me:
### Update

I've reviewed the existing components related to persistence:

- **Persistence.hpp**: This is an interface class responsible for defining how messages are stored and retrieved from a message queue.
  
- **PersistenceFactory.hpp**: A factory class that currently instantiates a `Persistence` object, limited to the SQLiteStorage for now.
  
- **SQLiteStorage**: A concrete implementation of the `Persistence` interface, handling the storage and retrieval of messages specifically within a SQLite database.
  
- **SQLiteManager**: A more generic SQL operation implementation, designed to provide broader database functionality.

I'm examining these classes to develop a structured plan moving forward.

jr0me:
### Update

Created branch and started doing a preparatory cleanup. Created outline for a common interface that can implement different relational db.

jr0me:
### Update

Worked on modifying AgentInfoPersistence to use the SQLiteManager.

jr0me:
### Update

AgentInfoPersistence now is fully implemented with SQLiteManager. Currently working on moving MultiTypeQueue and Persistence to use it too.

jr0me:
### Update

Made several fixes and separated what was ready for merging from the other developments. Branch was merged and issue is now on hold.

TomasTurina:
### Update

Resuming this issue. Will continue with the `multitype_queue`.

TomasTurina:
## Update

Several improvements in the `sqlite_manager`, `command_store` and `agent_info`.

Added support to `order by` and `limit` keywords in select queries.

TomasTurina:
## Update

Created `db_manager` component. Missing `multitype_queue` migration to use this component.

[...]

PR #496, opened 2025-01-09
Use a unique database manager and integrate it with the queue

TomasTurina:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/122|

## Description

WIP

## Tests

  - [ ] Linux
  - [ ] Windows
  - [ ] MAC OS X


src/agent/db_manager/src/sqlite_manager.cpp
```diff
@@ -0,0 +1,372 @@
...
+        while (query.executeStep())
+        {
+            int nColumns = query.getColumnCount();
+            Row queryFields;
+            queryFields.reserve(static_cast<size_t>(nColumns));
+            for (int i = 0; i < nColumns; i++)
+            {
+                queryFields.emplace_back(query.getColumn(i).getName(),
+                                         ColumnTypeFromSQLiteType(query.getColumn(i).getType()),
+                                         query.getColumn(i).getString());
```
LucioDonda, line 258: What if the value is not a string?

src/agent/db_manager/include/column.hpp
```diff
@@ -0,0 +1,98 @@
...
+class ColumnValue : public ColumnName
+{
+public:
+    /// @brief Constructor for defining a column with a specific value.
+    /// @param name The name of the column.
+    /// @param type The data type of the column.
+    /// @param value The value of the column.
+    ColumnValue(std::string name, const ColumnType type, std::string value)
+        : ColumnName(std::move(name), type)
+        , Value(std::move(value))
```
LucioDonda, line 87: Type can be TEXT INTEGER or REAL but while constructing ColumValue it only accepts string. Can this be an issue?

LucioDonda (commented):
just leaved some comments with column types doubts

src/agent/command_store/src/command_store.cpp
```diff
@@ -26,24 +26,38 @@ namespace command_store
...
+                columns.emplace_back("time", ColumnType::REAL, true, false, false);
+
+                try
+                {
+                    m_dataBase->CreateTable(COMMANDSTORE_TABLE_NAME, columns);
+                }
+                catch (std::exception& e)
+                {
+                    LogError("CreateTable operation failed: {}.", e.what());
+                }
```
jr0me: If the table can't be created due to an exception... the exception won't be propagated.

src/agent/agent_info/src/agent_info_persistance.cpp
```diff
@@ -60,30 +70,32 @@ void AgentInfoPersistance::CreateAgentInfoTable()
...
     }
 }
 
 void AgentInfoPersistance::CreateAgentGroupTable()
 {
     try
     {
-        const std::vector<Column> columns = {Column("id", ColumnType::INTEGER, true, true, true),
-                                             Column("name", ColumnType::TEXT, true, false)};
+        const Keys columns = {ColumnKey(AGENT_GROUP_ID_COLUMN_NAME, ColumnType::INTEGER, NOT_NULL | PRIMARY_KEY),
```
jr0me, line 90: Shouldn't this also be auto increment?

src/agent/command_store/include/command_store.hpp
```diff
@@ -1,27 +1,25 @@
 #pragma once
 
 #include <module_command/command_entry.hpp>
-#include <sqlite_manager.hpp>
+#include <persistence.hpp>
+#include <persistence_factory.hpp>
```
jr0me, line 5: Could the factory header go into the source file?

src/agent/multitype_queue/CMakeLists.txt
```diff
@@ -8,24 +8,23 @@ project(MultiTypeQueue)
...
 
 target_link_libraries(MultiTypeQueue
         PUBLIC
+        Persistence
         Config
         nlohmann_json::nlohmann_json
         Boost::asio
         Logger
         PRIVATE
-        SQLiteCpp
```
jr0me, line 28: 👌

src/agent/multitype_queue/include/storage.hpp
```diff
@@ -0,0 +1,119 @@
...
+#include <persistence.hpp>
+#include <persistence_factory.hpp>
+
+#include <nlohmann/json.hpp>
+
+#include <memory>
+#include <mutex>
+#include <string>
+
+/// @brief Implementation of the Persistence interface.
```
jr0me, line 12: Is this still the case? SQLiteStorage used to inherit from Persistence, but this one does not.

src/agent/multitype_queue/include/storage.hpp
```diff
@@ -0,0 +1,119 @@
+#pragma once
+
+#include <persistence.hpp>
+#include <persistence_factory.hpp>
```
jr0me, line 4: `Persistence` class can be forward declared (in other files may be too, but I didn't check). Then both header files can be moved to the source file, and linked privately.

If you have a unique pointer to the Persistence class and it is forward declared, I believe that you need to declare a default destructor explicitly.

src/agent/multitype_queue/src/storage.cpp
```diff
@@ -0,0 +1,342 @@
+#include <storage.hpp>
+
+#include <logger.hpp>
+
+#include "column.hpp"
```
jr0me, line 5: Column is from another cmake project, should use <>.

src/agent/persistence/include/column.hpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /// @brief The value of the column as a string
+    std::string Value;
+};
+
+using Names = std::vector<ColumnName>;
+using Keys = std::vector<ColumnKey>;
+using Row = std::vector<ColumnValue>;
+using Criteria = std::vector<ColumnValue>;
```
jr0me, line 96: Having this very generic names could pollute the global namespace. Previously in sqlite_manager they were using a namespace. I suggest enclosing this file's contents as well in one.

src/agent/persistence/include/sqlite_manager.hpp
```diff
@@ -0,0 +1,144 @@
+#pragma once
+
+#include <SQLiteCpp/SQLiteCpp.h>
```
jr0me, line 3: We should move this to the source file.

src/agent/persistence/CMakeLists.txt
```diff
@@ -3,21 +3,21 @@ cmake_minimum_required(VERSION 3.22)
...
+add_library(Persistence src/persistence_factory.cpp src/sqlite_manager.cpp)
 
 include(../../cmake/ConfigureTarget.cmake)
-configure_target(SQLiteManager)
+configure_target(Persistence)
 
-target_include_directories(SQLiteManager PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
-target_link_libraries(SQLiteManager PUBLIC SQLiteCpp PRIVATE fmt::fmt Logger)
+target_include_directories(Persistence PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_link_libraries(Persistence PUBLIC SQLiteCpp PRIVATE fmt::fmt Logger)
```
jr0me, line 20: SQLiteCpp can be private if usage is not exposed in header files.

jr0me (commented):
Some minor changes requested but looking good!

src/agent/db_manager/include/column.hpp
TomasTurina: This is done to have the data ready to be appended to the query, so it is not an issue.

src/agent/db_manager/src/sqlite_manager.cpp
TomasTurina: Sqlitecpp returns the representation of the value as a string, which can then be converted to any type, so this is not an issue.

src/agent/persistence/tests/sqlite_manager_test.cpp
```diff
@@ -0,0 +1,426 @@
...
+    std::cout << "---------- " << ret.size() << " rows returned. ----------" << '\n';
+    for (const auto& row : ret)
+    {
+        for (const auto& field : row)
+        {
+            std::cout << "[" << field.Name << ": " << field.Value << "]";
+        }
+        std::cout << '\n';
+    }
+}
```
LucioDonda, line 157: is neccesary to print the results?

LucioDonda (commented):
LGTM! just minor comment.

Issue #486, opened 2025-01-08
Adapt command arguments to parse a JSON object based on the command name


TomasTurina:
## Description

As specified in https://github.com/wazuh/wazuh-indexer/issues/618, to improve command interoperability it is necessary to change the command arguments to be an object instead of an array of strings.

Since each command has different parameters, we need to analyze each command, define the list of arguments it will accept (required and optional), and parse them from a JSON object instead of an array.

Example command:

```json
{
    "orders": [
        {
            "action": {
                "args": {},
                "name": "restart",
                "version": "5.0.0"
            },
            "source": "Users/Services",
            "document_id": "A8-62pMBBmC6Jrvqj9kW",
            "user": "Management API",
            "target": {
                "id": "d5b250c4-dfa1-4d94-827f-9f99210dbe6c",
                "type": "agent"
            }
        }
    ]
}
```

Currently, the agent only supports the `set-group` and `update-group` commands.

## Tasks

- [ ] Define the list of arguments required by the command `set-group`.
- [ ] Define the list of arguments required by the command `update-group`.
- [ ] Adapt the code to parse these arguments from the JSON object based on the command name.
- [ ] Adapt tests if necessary.
- [ ] Add commands description to documentation (`docs/ref`).

sdvendramini:
### Arguments required by the command `set-group`

```
{
    "orders":
    [
        {
            "action":
            {
                "args":
                {
                    "groups":
                    [
                        "group_1",
                        "group_2",
                        "...",
                        "group_N"
                    ],
                    
                },
                "name": "set-group",
                "version": "5.0.0"
            },
            "source": "Users/Services",
            "document_id": "A8-62pMBBmC6Jrvqj9kW",
            "user": "Management API",
            "target":
            {
                "id": "d5b250c4-dfa1-4d94-827f-9f99210dbe6c",
                "type": "agent"
            }
        }
    ]
}
```

### Arguments required by the command `set-group`

```
{
    "orders":
    [
        {
            "action":
            {
                "args":
                {},
                "name": "update-group",
                "version": "5.0.0"
            },
            "source": "Users/Services",
            "document_id": "A8-62pMBBmC6Jrvqj9kW",
            "user": "Management API",
            "target":
            {
                "id": "d5b250c4-dfa1-4d94-827f-9f99210dbe6c",
                "type": "agent"
            }
        }
    ]
}
```

sdvendramini:
## 2024/01/21

[...]

PR #531, opened 2025-01-22
Adapt command arguments to parse a JSON object based on the command name

sdvendramini:
|Related issue|
|---|
|https://github.com/wazuh/wazuh-agent/issues/486|

This PR introduces the necessary changes to parse a command with its arguments in JSON object format instead of an array, as previously used. To achieve this, the method of extracting commands from the queue has been modified, along with the function that validates commands. Additionally, a more detailed structure for each command's requirements has been defined, based on its name.

### DIFF
--- a/src/agent/centralized_configuration/src/centralized_configuration.cpp
+++ b/src/agent/centralized_configuration/src/centralized_configuration.cpp
@@ -19,7 +19,15 @@ namespace centralized_configuration
                 if (m_setGroupIdFunction && m_downloadGroupFilesFunction && m_validateFileFunction &&
                     m_reloadModulesFunction)
                 {
-                    groupIds = parameters.get<std::vector<std::string>>();
+                    groupIds = parameters.at("groups").get<std::vector<std::string>>();
+                    if (!std::all_of(
+                            groupIds.begin(), groupIds.end(), [](const std::string& id) { return !id.empty(); }))
+                    {
+                        LogWarn("Group name can not be an empty string.");
+                        co_return module_command::CommandExecutionResult {
+                            module_command::Status::FAILURE,
+                            "CentralizedConfiguration group set failed, a group name can not be an empty string."};
+                    }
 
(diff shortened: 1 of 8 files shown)


src/agent/command_handler/src/command_handler.cpp
```diff
@@ -2,10 +2,22 @@
...
 {
-    const std::unordered_map<std::string, std::pair<std::string, module_command::CommandExecutionMode>>
-        VALID_COMMANDS_MAP = {
-            {"set-group", {"CentralizedConfiguration", module_command::CommandExecutionMode::SYNC}},
-            {"update-group", {"CentralizedConfiguration", module_command::CommandExecutionMode::SYNC}}};
+    struct CommandDetails
+    {
+        std::string module;
+        module_command::CommandExecutionMode executionMode;
+        bool requiresArguments;
```
TomasTurina, line 9: We can remove this if we check the arguments vector size.

src/agent/centralized_configuration/src/centralized_configuration.cpp
```diff
@@ -19,7 +19,7 @@ namespace centralized_configuration
                 if (m_setGroupIdFunction && m_downloadGroupFilesFunction && m_validateFileFunction &&
                     m_reloadModulesFunction)
                 {
-                    groupIds = parameters.get<std::vector<std::string>>();
+                    groupIds = parameters.at("groups").get<std::vector<std::string>>();
```
TomasTurina, line 22: We may check again that `groupIds` is not empty.

TomasTurina, line 22: We may check again at this point that the key `groups` exists.

src/agent/command_handler/src/command_handler.cpp
sdvendramini: Done!

TomasTurina (approved):

PR #543, opened 2025-01-24
Manage additional dependencies with vcpkg

jr0me:
## Description

This PR adds additional dependencies that were previously installed system-wide to vcpkg.

The main changes come to the `librpm` port overlay. There were some missing dependencies (libmagic, bzip2, etc) and some missing configuration.

This configuration is needed for our custom port to use the dependencies fetched by vcpkg, furthermore, there were some inconsistencies in the naming of library files that needed to be resolved manually (popt, bz2, Lua).


BUILD.md
```diff
@@ -1,17 +1,5 @@
 # Build Instructions
 
-The following dependencies are required for this project:
```
jr0me, line 3: I removed this because the dependencies are explicitly listed below.

TomasTurina (approved):

TomasTurina (approved):

PR #627, opened 2025-02-25
Add support for restart command in Windows Agent

aritosteles:
## Description

This PR enables the Wazuh agent to support a command for self-restart in Windows.

## Proposed Changes

- Devise a strategy to restart the Agent when running as a service. Windows Service Control does not offer a restart command like launchd or systemd do. This is problematic since the service cannot restart itself using a simple sequence of stop/start because it is stopped before it has a chance to start another instance of itself. The desired functionality can be achieved by leveraging Powershell command `Restart-Service`, running through a call to Win32 function CreateProcess().
- Devise a strategyto restart the Agent when running in the foreground. This presents a problem similar to the one described in the previous case because we are not running a separate process to do the restart. The problem can be solved by a call to 'cmd' before stopping the Agent. The instance of cmd will delay for some time to allow the running Agent to gracefully shutdown before starting a new instance.

### Results and Evidence

- Agent restarted while running as a service:

image

Full sequence of events:

[event log.txt](https://github.com/user-attachments/files/18984242/event.log.txt)

- Agent restarted while running in the foreground:

image

### Artifacts Affected

- Windows executable

### Configuration Changes

None

### Documentation Updates

None

## Review Checklist

- [x] Code changes reviewed
- [x] Relevant evidence provided
- [x] Configuration changes documented
- [x] Developer documentation reflects the changes
- [x] Meets requirements and/or definition of done
- [x] No unresolved dependencies with other issues
- [ ] ...

### DIFF
--- a/src/agent/restart_handler/CMakeLists.txt
+++ b/src/agent/restart_handler/CMakeLists.txt
@@ -8,17 +8,18 @@ set_common_settings()
 find_package(Boost REQUIRED COMPONENTS asio)
 
 if(WIN32)
-    set(SOURCES src/restart_handler_win.cpp)
+    set(SOURCES src/restart_handler.cpp src/restart_handler_win.cpp)
 elseif(APPLE)
-    set(SOURCES src/restart_handler_unix.cpp src/restart_handler_mac.cpp)
+    set(SOURCES src/restart_handler.cpp src/restart_handler_unix.cpp src/restart_handler_mac.cpp)
 else()
-    set(SOURCES src/restart_handler_unix.cpp src/restart_handler_lin.cpp)
+    set(SOURCES src/restart_handler.cpp src/restart_handler_unix.cpp src/restart_handler_lin.cpp)
 endif()
 
 add_library(RestartHandler ${SOURCES})
 target_include_directories(RestartHandler
                             PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
                             PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+
 target_link_libraries(RestartHandler PUBLIC Boost::asio CommandEntry PRIVATE Logger)
 
 include(../../cmake/ConfigureTarget.cmake)
--- a/src/agent/restart_handler/src/restart_handler_win.cpp
+++ b/src/agent/restart_handler/src/restart_handler_win.cpp
@@ -1,16 +1,123 @@
+#include <restart_handler_win.hpp>
+
 #include <logger.hpp>
-#include <restart_handler.hpp>
+#include <shlobj_core.h>
+
+#include <algorithm>
+#include <fmt/format.h>
 
 namespace restart_handler
 {
 
     std::vector<char*> RestartHandler::startupCmdLineArgs;
 
-    boost::asio::awaitable<module_command::CommandExecutionResult> RestartHandler::RestartAgent()
+    boost::asio::awaitable<module_command::CommandExecutionResult> RestartForeground()
     {
-        // TODO
-        co_return module_command::CommandExecutionResult {module_command::Status::FAILURE,
-                                                          "RestartHandler is not implemented yet"};
+        STARTUPINFO si = {0};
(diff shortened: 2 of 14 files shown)


jr0me (commented):
Will this only work if windows has powershell? I ask because I'm not sure that it is always available.

src/agent/restart_handler/src/restart_handler_win.cpp
```diff
@@ -1,16 +1,69 @@
-#include <logger.hpp>
 #include <restart_handler.hpp>
 
+#include "../../src/windows/windows_service.hpp"
```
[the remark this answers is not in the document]
jr0me: Looks like my comment is outdated.

src/agent/restart_handler/CMakeLists.txt
```diff
@@ -18,7 +18,9 @@ endif()
 add_library(RestartHandler ${SOURCES})
 target_include_directories(RestartHandler
                             PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
-                            PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src)
+                            PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src
+                            ${CMAKE_CURRENT_SOURCE_DIR}/../src/windows)
```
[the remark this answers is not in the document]
jr0me: Outdated comment.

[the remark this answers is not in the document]
aritosteles: Done

aritosteles:
> Will this only work if windows has powershell? I ask because I'm not sure that it is always available.

Windows Powershell has been bundled since Windows XP SP3 and can be installed on earlier versions of XP. Should this be a concern?

jr0me:
> Windows Powershell has been bundled since Windows XP SP3 and can be installed on earlier versions of XP. Should this be a concern?

Not necessarily, but if it's not bundled in any of the Windows versions we support, it should be added to the requirements list or to the documentation.

TomasTurina (approved):

src/agent/restart_handler/src/restart_handler_win.cpp
```diff
@@ -1,16 +1,125 @@
...
+        co_return module_command::CommandExecutionResult {module_command::Status::IN_PROGRESS,
+                                                          "Pending restart execution"};
     }
 
+    std::string GetPowerShellPath()
+    {
+        auto is64bits = []() -> bool
+        {
+            SYSTEM_INFO systemInfo;
+            GetNativeSystemInfo(&systemInfo); // Get native system architecture
```
jr0me, line 55: I think that we can remove the comment since it repeats the name of the function called

src/agent/restart_handler/src/restart_handler_win.cpp
```diff
@@ -1,16 +1,125 @@
...
+            {
+                return "powershell";
+            }
+
+            std::string result(size, 0);
+            WideCharToMultiByte(CP_UTF8, 0, powershellPath.c_str(), -1, &result[0], size, nullptr, nullptr);
+            return result.c_str(); // returning c_str() removes the trailing null char
+        }
+
+        return "powershell";
```
jr0me, line 84: So we don't repeat the `return "powershell"`

You can do

```c++
if (size > 0)
{
            std::string result(size, 0);
            WideCharToMultiByte(CP_UTF8, 0, powershellPath.c_str(), -1, &result[0], size, nullptr, nullptr);
            return result.c_str(); // returning c_str() removes the trailing 
}

jr0me (approved):
LGTM!

src/agent/restart_handler/src/restart_handler_win.cpp
aritosteles: Done

src/agent/restart_handler/src/restart_handler_win.cpp
aritosteles: Removed

PR #701, opened 2025-03-26
Update event id and event name in windows

cborla:
## Description

The Windows agent logs events in the Windows event store (official name?). These logs are accessible via the Event Viewer.

However, the logs appear with ID=0 and contain unclear descriptions, making log interpretation difficult.

We should clarify event logging to ensure they are displayed correctly in the Event Viewer. 

## Proposed Changes

Changing the event ID to 0.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided
- [ ] Configuration changes documented
- [ ] Developer documentation reflects the changes
- [ ] Meets requirements and/or definition of done
- [ ] No unresolved dependencies with other issues


packages/windows/postinstall.ps1
```diff
@@ -73,4 +73,23 @@ if (-not (Get-Service -Name $serviceName -ErrorAction SilentlyContinue)) {
...
+    try {
+        Write-Host "Registering event log source: $sourceName"
+        New-EventLog -LogName $logName -Source $sourceName
+        Write-Host "Event log source registered successfully."
+    } catch {
+        Write-Host "Failed to register event log source: $_"
+        exit 1
+    }
+} else {
+    Write-Host "Event log source '$sourceName' is already registered."
```
LucioDonda, line 92: can this be an issue?

src/common/logger/src/logger_win.cpp
```diff
@@ -5,12 +5,14 @@
 
 #include <memory>
 
+#define EVENT_ID 1000
```
LucioDonda, line 8: this ID will remain fixed no matter what events of the agents are being logged?

LucioDonda (commented):
Besides those doubts is there a chance of opening of a new issue with the enhancement of the windows login making use of several IDs depending on the events? 
can this be used in other modules of the agent?

packages/windows/postinstall.ps1
cborla: No, if the name wazuh-agent is already registered in Eventlog there is no need to register it again, the messages will be displayed correctly anyway.

src/common/logger/src/logger_win.cpp
cborla: Initially, the ID will be unique for all the agent events.

LucioDonda (approved):
LGTM

PR #788, opened 2025-05-15
Centralize rule evaluation timers into SCA

jr0me:
## Review Checklist

- [ ] Code changes reviewed
- [ ] Relevant evidence provided
- [ ] Configuration changes documented
- [ ] Developer documentation reflects the changes
- [ ] Meets requirements and/or definition of done
- [ ] No unresolved dependencies with other issues
- [ ] ...


src/agent/task_manager/src/task_manager.cpp
```diff
@@ -13,25 +13,55 @@ TaskManager::~TaskManager()
...
+    {
+        const std::lock_guard<std::mutex> lock(m_mutex);
+
+        if (!m_work)
+        {
+            m_work = std::make_unique<boost::asio::executor_work_guard<boost::asio::io_context::executor_type>>(
+                boost::asio::make_work_guard(m_ioContext));
+        }
+
+        if (m_ioContext.stopped())
```
TomasTurina, line 54: Shouldn't this be at the beginning or is the the same?

src/agent/task_manager/src/task_manager.cpp
TomasTurina, line 1: Please upload evidence of all commands execution, specially those that reload config and/or modules.

jr0me: Updated with logs from macos and linux.

src/modules/logcollector/tests/unit/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ endif()
 file(GLOB LOGCOLLECTOR_TEST_SOURCES *_test.cpp)
 file(GLOB UNIX_TEST_SOURCES journald_reader/*.cpp file_reader/*_unix_test.cpp)
 file(GLOB MACOS_TEST_SOURCES macos_reader/*.cpp file_reader/*_unix_test.cpp)
-file(GLOB WIN_TEST_SOURCES winevt_reader/*.cpp file_reader/*_win_test.cpp)
+file(GLOB WIN_TEST_SOURCES file_reader/*_win_test.cpp)
```
TomasTurina, line 9: What about these tests?

src/modules/sca/include/sca.hpp
```diff
@@ -95,6 +96,6 @@ class SecurityConfigurationAssessment : public IModule
     /// @brief List of policies
     std::vector<std::unique_ptr<ISCAPolicy>> m_policies;
 
-    /// @brief Boost ASIO context
-    std::unique_ptr<boost::asio::io_context> m_ioContext;
+    /// @brief Task manager for managing tasks
+    std::unique_ptr<TaskManager> m_taskManager;
```
TomasTurina, line 100: Just wondering why we use a TaskManager pointer in SCA but not in other modules.

src/agent/task_manager/src/task_manager.cpp
TomasTurina: But I think that value will always be 0, right? Why do we need it?
```
    ++m_numAdditionalThreads;
     m_ioContext.run();
     --m_numAdditionalThreads;
```

src/agent/task_manager/include/task_manager.hpp
```diff
@@ -48,7 +66,10 @@ class TaskManager : public ITaskManager<boost::asio::awaitable<void>>
     std::vector<std::thread> m_threads;
 
     /// @brief Number of enqueued threads
-    std::atomic<size_t> m_numEnqueuedThreads = 0;
+    std::atomic<size_t> m_numEnqueuedThreadTasks = 0;
+
+    /// @brief Number of additional threads (if any) that are not part of the thread pool
+    std::atomic<size_t> m_numAdditionalThreads = 0;
```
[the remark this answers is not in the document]
TomasTurina: But I think that value will always be 0, right? Why do we need it?

```c
     ++m_numAdditionalThreads;
     m_ioContext.run();
     --m_numAdditionalThreads;
```

src/modules/logcollector/tests/unit/CMakeLists.txt
jr0me: Winevt? I removed them because they didn't test the reader.

src/modules/sca/include/sca.hpp
jr0me: No real reason (before it was a unique pointer of io context), maybe we should inject it as a dependency to be able to test SCA and LogCollector without the TaskManager. But right now I didn't focus in adding tests.

src/modules/logcollector/tests/unit/CMakeLists.txt
jr0me: The FileReader tests seem to be testing FileReader functionality, so I left them.

src/agent/task_manager/include/task_manager.hpp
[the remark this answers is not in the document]
jr0me: This call is blocking: `m_ioContext.run()`. So if many threads would call it, the number would increase until `Stop` is called.

PR #851, opened 2025-09-09
feat(#25587): add CephFS type to skip_nfs flag

### LINKED ISSUE #25587: Add CephFS type to skip_nfs flag
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| v4.7.2-40711 | rootcheck(and others)| Agent | APT repo Package | Ubuntu 22.04.3 LTS |

I completely understand why nfs and cifs have an option to be skipped. That is because it's a remote filesystem and multiple or agants scanning the same files is not a good idea. However NFS and CIFS are not the only ones. 

For example: 
I have CephFS volumes mounted on multiple (hundreds) pods in multiple kubernetes clusters with "ReadWriteMany" mode. Meaning they all have that mounted at the same time. So agent being deployed on all nodes just traverses everything under /var/lib with rootcheck.

`/var/lib/kubelet/pods/<UID>/volumes/kubernetes.io~csi/<volume_name>/mount` is mounted multiple times even on the same node.

`lsof` shows a tone of open operations typed `DIR`  I don't want to get into cephfs internals, but this in ends up in slow operations or oversized cache between client and ceph metadata service, since the same operations are coming from a lot of clients and the same time.

Mounting is done from kernel, not fuse. 

Since "yet-another-hardcoded-value" does not look like the best possible solution, my suggestion is to make this configurable somehow. Like there's the skipFS function:
https://github.com/wazuh/wazuh-agent/blob/master/src/common/fs_op/src/fs_op.c#L96
I think this could be configurable. Though the values of FS type are predefined, I don't believe there isn't another way to detect FS type.

stevenc99:
[ceph-csi](https://github.com/ceph/ceph-csi) for Kubernetes mounts CephFS filesystems under /var/lib/kubelet/plugins/kubernetes.io/csi/cephfs.csi.ceph.com/*/globalmount
 
Wazuh rootcheck scans include /var/lib by default
https://documentation.wazuh.com/current/user-manual/reference/ossec-conf/rootcheck.html?utm_source=chatgpt.com#readall
 
As a result, Wazuh on Kubernetes hosts will recurse into those network filesystems, which is typically not desired, as they may be mounted by multiple hosts at the same time, as well as potentially causing performance issues.
 
I suggest that skip_nfs include CephFS (but not necessarily Ceph RBD) in addition to CIFS and NFS.

I can confirm that this patch worked as intended for the Linux systems that I'm administrator for. This was confirmed by manually triggering a rootcheck via
`/var/ossec/bin/agent_control -r`
and observing the scanned paths on the monitored hosts with
`strace -f -e trace=lstat -p $(pidof wazuh-syscheckd)`
 
Closes https://github.com/wazuh/wazuh/issues/25587


vikman90:
@stevenc99 thanks for the contribution.

As mentioned in https://github.com/wazuh/wazuh/pull/31846, we’ll proceed with handling this scenario via Rootcheck `<exclude>` path configuration rather than adding filesystem-type-based exclusions. For that reason, we’re closing this PR as well.

Appreciate your time and effort.

Best regards

